PR #24, opened 2016-01-20
Capture only one Middleware Layer per request

cschneid:
This will limit us to summary metrics only, instead of per-middleware
timings.  But we already had to limit per-middleware timing due to
overhead concerns.

The old code still is here, in the MiddlewareDetailed class, but isn't
enabled.  I imagine that we'll get a flag in the config to record more
detailed metrics for the applications that need that detail to debug
performance issues.

By only recording one thing, we can remove the special casing around
storing the middleware metrics, which we put in to minimize the space
required to store SlowTransaction records in the layaway file.

This also fixes a bug in the LayerDepthFirstWalker class, where it never
yielded the root_layer itself to the callback.  That wasn't very relevant
when it was a rails built-in middleware that took approximately zero
time, but it became pronounced when doing development here and it was
skipping the entire middleware layer, which happens to be the
root_layer since it comes first in a request.


itsderek23:
Testing this out - couple of UI things we'll want to adjust:

slow_stream___scout

PR #26, opened 2016-01-26
Lazy sanitization of SQL, and a couple of performance improvements

itsderek23:
Previously, we were sanitizing SQL queries as layers were added. Now, we sanitize only where it is needed: `LayerSlowTransactionConverter`. This results in large performance gains:

* overhead benchmarks drop from 8.1% to 3.2%
* stackprof samples for `ScoutApm::Utils::SqlSanitizer#to_s_postgres` drop from 1.2% to 0%

This also addresses a couple other higher-sample % numbers from stackprof output:

* Memo-izing fetching the database driver
* Using strip! vs. gsub!

This needs to run on production environments first before merging.

### DIFF
--- a/lib/scout_apm/utils/sql_sanitizer.rb
+++ b/lib/scout_apm/utils/sql_sanitizer.rb
@@ -40,7 +40,7 @@ module ScoutApm
         sql.gsub!(PSQL_REMOVE_INTEGERS, '?')
         sql.gsub!(PSQL_IN_CLAUSE, 'IN (?)')
         sql.gsub!(MULTIPLE_SPACES, ' ')
-        sql.gsub!(TRAILING_SPACES, '')
+        sql.strip!
         sql
       end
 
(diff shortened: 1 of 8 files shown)


lib/scout_apm/utils/sql_sanitizer.rb
```diff
@@ -57,7 +60,8 @@ def to_s_sqlite
         sql.gsub!(SQLITE_REMOVE_STRINGS, '?')
         sql.gsub!(SQLITE_REMOVE_INTEGERS, '?')
```
cschneid: If you want to get fancy, you can do this in a single regex pass by doing:

```
sql.gsub!(SQLITE_REMOVE_STRINGS.union(SQL_REMOVE_INTEGERS), '?')
```

No clue if its faster, but it'd be one less call to `gsub!`

cschneid:
:+1:

itsderek23:
This has been running on prod for a while and looking good. Bumped version to 1.2.12 and updated changelog.

itsderek23:
> Bumped version to 1.2.12 and updated changelog.

Actually 1.2.13. 1.2.12 was already released.

PR #28, opened 2016-01-29
uri reporting config option (path or fullpath)

andre:
For customers that don't want URL params propagated into Scout. Default is fullpath.


cschneid:
:+1:

PR #31, opened 2016-03-01
Add HttpClient instrumentation

cschneid:


cschneid:
screenshot 2016-03-01 14 34 29

PR #32, opened 2016-03-02
Add support for URI urls

nicdal:
Add support for using URI urls instead of only strings. Will get tools like Savon / httpi to work.


cschneid:
Sorry about that! I tested it with HTTPClient directly, passing in a string, so it didn't come up.

Merged and will be released as 1.4.3 in a few minutes.

PR #35, opened 2016-03-15
Added MetricMeta#kind attribute to support objects

itsderek23:
This adds a `MetricMeta#kind` attribute, which is used to identify object allocation metrics. 

kind = "objects" || nil.

Requires a sever-side change.


cschneid:
This will bleed into json encoding, node-agg's aggregation of metrics, then the metrics ingest.

andre:
I don't have the full understanding of why we need `kind` in addition to `type`. Can we talk through on Monday?

itsderek23:
Sure.

PR #39, opened 2016-04-19
Background Job Monitoring - DO NOT MERGE

itsderek23:
1.5.0.pre version of the agent.


lib/scout_apm/background_job_integrations/sidekiq.rb
```diff
@@ -12,48 +12,60 @@ def present?
...
         begin
           yield
         rescue
           req.error!
           raise
-        ensure
-          req.stop_layer
         end
+      ensure
+        req.stop_layer # Job
```
itsderek23: How come we call `req#stop_layer` twice?

cschneid: Because we start two layers. `Queue` and `Job`.

PR #48, opened 2016-06-17
Dev mode

andre:
Initial implementation: middleware, config changes, isolated serialization of just one trace, etc.

Still has some development convenience stuff in it:

* The middleware automatically re-loads itself each request
* the XMLHttpRequest instrumentation is verbose (and it may ultimately live someplace else), and has a debug statement.


lib/scout_apm/instant/middleware.rb
```diff
@@ -0,0 +1,116 @@
...
+      def add_to_body(content)
+        @to_add_to_body << content
+      end
+
+      def res
+        i = @html.index("</body>")
+        @html = @html.insert(i, @to_add_to_body.join("")) if i
+        i = @html.index("</head>")
+        @html = @html.insert(i, @to_add_to_head.join("")) if i
+        @html
```
cschneid: Is it intentional to silently fail to add stuff to the page if you don't find the indexes?

lib/scout_apm/instant/middleware.rb
```diff
@@ -0,0 +1,116 @@
...
+                end
+              else
+                [status, headers, response]
+              end
+            else
+              [status, headers, response]
+            end
+          else
+            [status, headers, response]
+          end
```
cschneid: Can you invert some of these ifs, so it's `return [status, headers, response] if CONDITION`?  Would keep the nesting down.

lib/scout_apm/instant/middleware.rb
```diff
@@ -0,0 +1,116 @@
...
+      #
+      # def call(env)
+      # end
+
+      def initialize(*args)
+        @args = args
+      end
+
+      def call(env)
+        "#{self.class}::Logic".constantize.new(*@args).call(env)
```
cschneid: What's going on here? Why not just say `Middleware::Logic.new().call(env)`? Or even fully qualified name. Why the constantize?

lib/scout_apm/instant/middleware.rb
andre: Yeah, it keeps us from adding to anything that doesn't have a body & head

lib/scout_apm/instant/middleware.rb
andre: It's forcing the middleware code to reload. I'll take this out before general release.

lib/scout_apm/instant/middleware.rb
andre: probably. I'll do a general code cleanup at some point.

itsderek23:
👏

PR #58, opened 2016-08-15
Use 1.8 compatible hash syntax in Instant::Middleware

conarro:
Ran into this when testing out the latest gem version locally on Ruby 1.8.7.


cschneid:
Thank you!

PR #61, opened 2016-08-21
Better security-by-default for Transaction Trace URIs

itsderek23:
Applying `filtered_parameters` settings to Rails 3+ trace uris. For Rails 2, just reporting the path.

These are better security-by-default settings for transaction trace uris. Example formatting of URIs for Rails 3+ reported to Scout:

`/users?password=secret&limit=10 => /users?password=[FILTERED]&limit=10`


andre:
Looks good.

andre:
I'll update help

PR #67, opened 2016-08-25
This refines what it means to be "in-app" when it comes to backtraces

cschneid:
If you're in the Rails root, and inside lib/ app/ or config/, then you
are.  This prevents directories like vendor/ from showing up in the UI
as code the user owns directly.

### DIFF
--- a/lib/scout_apm/utils/backtrace_parser.rb
+++ b/lib/scout_apm/utils/backtrace_parser.rb
@@ -14,7 +14,10 @@ module ScoutApm
       def initialize(call_stack, root=ScoutApm::Environment.instance.root)
         @call_stack = call_stack
         # We can't use a constant as it'd be too early to fetch environment info
-        @@app_dir_regex = %r|#{root}/(.*)|
+        #
+        # This regex looks for files under the app root, inside lib/, app/, and
+        # config/ dirs, and captures the path under root.
+        @@app_dir_regex = %r|#{root}/((?:lib/|app/|config/).*)|
       end
 
       def call
--- a/lib/scout_apm/utils/backtrace_parser.rb
+++ b/lib/scout_apm/utils/backtrace_parser.rb
@@ -17,7 +17,7 @@ module ScoutApm
         #
         # This regex looks for files under the app root, inside lib/, app/, and
         # config/ dirs, and captures the path under root.
-        @@app_dir_regex = %r|#{root}/((?:lib/|app/|config/).*)|
+        @@app_dir_regex = %r[#{root}/((?:lib/|app/|config/).*)]
       end
 
       def call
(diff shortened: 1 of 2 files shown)


lib/scout_apm/utils/backtrace_parser.rb
```diff
@@ -14,7 +14,10 @@ class BacktraceParser
       def initialize(call_stack, root=ScoutApm::Environment.instance.root)
         @call_stack = call_stack
         # We can't use a constant as it'd be too early to fetch environment info
-        @@app_dir_regex = %r|#{root}/(.*)|
+        #
+        # This regex looks for files under the app root, inside lib/, app/, and
+        # config/ dirs, and captures the path under root.
+        @@app_dir_regex = %r|#{root}/((?:lib/|app/|config/).*)|
```
itsderek23: Does it make sense to do a quick benchmark comparing the regex performance of this change since its a frequent call?

cschneid:
Quick benchmark:

```
› ruby regex_bench.rb
       user     system      total        real
old:  0.820000   0.010000   0.830000 (  0.829205)
new:  0.590000   0.010000   0.600000 (  0.609891)
```

(With code:)

```ruby
require 'benchmark'

n = 50000

root = "/Users/cschneid/foo"

array = [
  "#{root}/vendor/library_code/bar_calculator.rb",
  "#{root}/vendor/library_code/bar_calculator.rb",
  "#{root}/vendor/library_code/bar_calculator.rb",
  "#{root}/app/controllers/foo_controller.rb",
  "#{root}/app/controllers/foo_controller.rb",
  "#{root}/app/controllers/foo_controller.rb",
  "#{root}/app/controllers/foo_controller.rb",
  "#{root}/vendor/library_code/bar_calculator.rb",
  "#{root}/vendor/library_code/bar_calculator.rb",
  "#{root}/vendor/library_code/bar_calculator.rb",
  "#{root}/vendor/library_code/bar_calculator.rb",
  "#{root}/app/controllers/foo_controller.rb",
  "#{root}/app/controllers/foo_controller.rb",
  "#{root}/app/controllers/foo_controller.rb",
]

old_app_dir_regex = %r[#{root}/.*]
new_app_dir_regex = %r[#{root}/((?:lib/|app/|config/).*)]

Benchmark.bm do |x|
  x.report("old:") {
    n.times do
      array.each { |a| a.match(old_app_dir_regex) }
    end
  }

  x.report("new:") {
    n.times do
      array.each { |a| a.match(new_app_dir_regex) }
    end
  }
end
```

cschneid:
In total, it's doing 700k regex matches for each benchmark result.  And completing them in less than a second.

Run against ruby 2.2.2

itsderek23:
👍

itsderek23:
I approve this PR. 

:gavel

PR #74, opened 2016-09-01
Send full histograms up with the payload

cschneid:
This is more mathematically accurate merging of many payloads coming in for
95th percentile. We're dealing with approximate histograms either way, but this
waits till the last moment before pulling a single number out.  It will better
reflect the 95th% of the entire pool of requests, rather than an average of 95%
across each server.


dlanderson:
@cschneid can you post the maths again for how this will impact the size of the payload?

lib/scout_apm/config.rb
```diff
@@ -159,7 +159,8 @@ class ConfigDefaults
         'enable_background_jobs' => true,
         'ignore'                 => [],
         'dev_trace'              => false,
-        'profile'                => true # for scoutprof
+        'profile'                => true, # for scoutprof
+        'compress_payload'       => false,
```
dlanderson: Why false by default?

cschneid: It will be true for release. for the moment I just don't want to default to breaking things if it does infact break things.

Issue #76, opened 2016-09-13
Isn't working with mongoid


walterjsilva:
Hi, I'm trying to use the scout to monitor my rails app, but when beginning to track pops up a follow error:
```
NoMethodError: undefined method `filter' for #<Mongo::Collection::View:0x007f6c315731c8>

backtrace:
(eval):8 in first_with_scout_instruments
vendor/bundle/ruby/2.2.0/gems/mongoid-5.0.0/lib/mongoid/contextual.rb:20 in first
vendor/bundle/ruby/2.2.0/gems/orm_adapter-0.5.0/lib/orm_adapter/adapters/mongoid.rb:22 in get
vendor/bundle/ruby/2.2.0/gems/devise-3.5.2/lib/devise/models/authenticatable.rb:208 in serialize_from_session
vendor/bundle/ruby/2.2.0/gems/devise-3.5.2/lib/devise.rb:455 in block (2 levels) in configure_warden!
vendor/bundle/ruby/2.2.0/gems/warden-1.2.3/lib/warden/session_serializer.rb:34 in fetch
vendor/bundle/ruby/2.2.0/gems/warden-1.2.3/lib/warden/proxy.rb:212 in user
vendor/bundle/ruby/2.2.0/gems/warden-1.2.3/lib/warden/proxy.rb:318 in _perform_authentication
vendor/bundle/ruby/2.2.0/gems/warden-1.2.3/lib/warden/proxy.rb:104 in authenticate
vendor/bundle/ruby/2.2.0/gems/warden-1.2.3/lib/warden/proxy.rb:114 in authenticate?
vendor/bundle/ruby/2.2.0/gems/devise-3.5.2/app/controllers/devise_controller.rb:103 in require_no_authentication
vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.4/lib/active_support/callbacks.rb:432 in block in make_lambda
vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.4/lib/active_support/callbacks.rb:145 in call
vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.4/lib/active_support/callbacks.rb:145 in block in halting_and_conditional
vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.4/lib/active_support/callbacks.rb:504 in call
vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.4/lib/active_support/callbacks.rb:504 in block in call
[...]

PR #77, opened 2016-09-13
Fix incompatibility with Mongo gem 2.1

cschneid:
Mongoid 5 requires Mongo ~> 2.1. In development, we tested against 2.3
(current), but a customer ran into an issue where 2.1 was causing
issues.  After investigation, I found that between Mongo gem version 2.1
and 2.3, the name of the accessor that stores the active filters was
changed. This change is meant to accomodate both newer & older form.


cschneid:
Fixes #76

PR #79, opened 2016-10-24
Handle Addressable::URI in addition to URI

mrloop:
I've application code using Addressable::URI thus

    addressable_uri = Addressable::URI.parse(uri)
    Net::HTTP.get(addressable_uri)

This break with an exception being thrown by
`lib/scout_apm/instruments/net_http.rb`. One possilbe solution would be

    addressable_uri = URI(uri)
    Net::HTTP.get(addressable_uri)

But I'm on ruby 2.1.9 and have to handle subdomains with underscores,
see https://bugs.ruby-lang.org/issues/9974. So until I can update ruby >
2.1 my application code breaks with scout_apm.

This commit make scout_apm request profiling usable with both
Addressable::URI and URI.

### DIFF
--- a/lib/scout_apm/instruments/net_http.rb
+++ b/lib/scout_apm/instruments/net_http.rb
@@ -22,10 +22,15 @@ module ScoutApm
             include ScoutApm::Tracer
 
             def request_with_scout_instruments(*args,&block)
-              url = (@address + args.first.path.split('?').first)[0..99]
-              self.class.instrument("HTTP", "request", :desc => url) do
+              self.class.instrument("HTTP", "request", :desc => request_scout_description(args.first)) do
                 request_without_scout_instruments(*args, &block)
               end
+            end
+
+            def request_scout_description(req)
+              path = req.path
+              path = path.path if path.respond_to?(:path)
+              (@address + path.split('?').first)[0..99]
             end
 
             alias request_without_scout_instruments request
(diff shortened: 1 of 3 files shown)


lib/scout_apm/instruments/net_http.rb
```diff
@@ -22,12 +22,17 @@ def install
...
-              url = (@address + args.first.path.split('?').first)[0..99]
-              self.class.instrument("HTTP", "request", :desc => url) do
+              self.class.instrument("HTTP", "request", :desc => request_scout_description(args.first)) do
                 request_without_scout_instruments(*args, &block)
               end
             end
 
+            def request_scout_description(req)
+              path = req.path
+              path = path.path if path.respond_to?(:path)
```
cschneid: This bit is new - where `.path` gets called a second time.  How does that happen? Is it part of `Addressable::URI`?

cschneid:
Thank you for the PR! Just the one question, and I can get this further tested & merged in.

lib/scout_apm/instruments/net_http.rb
mrloop: Yes, see https://github.com/ruby/ruby/blob/v2_1_9/lib/net/http/generic_request.rb#L23 the `Addressable::URI` is assigned to path. Calling `path` returns the `Addressable::URI` not the expected string.

mrloop: Ideally `GenericRequest` wouldn't strictly check if it was a `URI` but rather use whatever it it had the same methods.

cschneid:
Thank you for the quick reply! I'll get this merged in, and make a release soon (probably a few days from now).

Feel free to point your Gemfile at our `master` until then, or you can continue using your own fork.

PR #80, opened 2016-11-02
Don't create unnecessary variables in Layer#initialize

cschneid:
Most Layer objects are leaf nodes (no children), with no annotations, so
avoid creating the empty data structures to hold data that will likely
never arrive.

This will lower the CPU overhead a small amount (no need to call then destroy an
extra set of objects), and will chip away at any memory overhead as
well, especially on long-running requests that create many Layer objects

### DIFF
--- a/lib/scout_apm/layer.rb
+++ b/lib/scout_apm/layer.rb
@@ -13,13 +13,17 @@ module ScoutApm
     #   instrumentation for an example of how this is useful
     attr_accessor :name
 
-    # An array of children layers, in call order.
+    # An array of children layers
     # For instance, if we are in a middleware, there will likely be only a single
     # child, which is another middleware.  In a Controller, we may have a handful
     # of children: [ActiveRecord, ActiveRecord, View, HTTP Call].
     #
     # This useful to get actual time spent in this layer vs. children time
-    attr_reader :children
+    #
+    # TODO: Check callers for compatibility w/ nil to avoid making an empty array
+    def children
+      @children || []
+    end
 
     # Time objects recording the start & stop times of this layer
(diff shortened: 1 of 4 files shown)


itsderek23:
Passes the eye test and less resource usage is always a 👏 .

Deploy the branch to our apps and ensure we're good?

lib/scout_apm/layer.rb
```diff
@@ -13,13 +13,17 @@ class Layer
...
-    # An array of children layers, in call order.
+    # An array of children layers
     # For instance, if we are in a middleware, there will likely be only a single
     # child, which is another middleware.  In a Controller, we may have a handful
     # of children: [ActiveRecord, ActiveRecord, View, HTTP Call].
     #
     # This useful to get actual time spent in this layer vs. children time
-    attr_reader :children
+    #
+    # TODO: Check callers for compatibility w/ nil to avoid making an empty array
```
dlanderson: What's the implication if this happens?

cschneid: It's better to just never make the empty array, since 99% of the time, a caller just calls  `.each` on it, which ends up being a noop, but I didn't want to spend the time right now to find and fix all callers.

No harm in making the array, just an operation that doesn't add to anything.  This is still a benefit because the Layer-at-rest won't hold onto an empty array.

dlanderson: Ah, the comment wasn't clear to me when I read it. Makes sense

cschneid:
Fixed a small typo that we found when we canary-deployed to APM. otherwise, everything is happy.

PR #81, opened 2016-11-04
Add and wireup a LayerChildrenSet to cap the number of children for a layer

cschneid:
This lets us cap the number of children any given layer can accumulate,
then it starts combining them together.

The limit is per type (HTTP, ActiveRecord, View...), and currently is at
1000.   That number is arbitrary, and not based on any research.  Likely it
will change or get larger as we figure out the right balance.

The reason for this is to limit the number of Layer objects we store in ram
over the course of a request. In long running jobs in particular, the
number of layers encountered can get very very large, and even though they
are pretty small objects on their own, the weight of them can add up.  This
approach provides a safety valve to prevent our memory use going up over a
long trace.

Anything going over the limit gets bucketed into a single `HTTP/Merged`
bucket, which shows up as its own line-item in the trace, so overall times
and counts are not lost.

### DIFF
--- a/lib/scout_apm/merged_layer.rb
+++ b/lib/scout_apm/merged_layer.rb
@@ -1,124 +1,122 @@
 module ScoutApm
-  module Layers
-    # A MergedLayer is a lossy-compression approach to fall back on once we max out
-    # the number of detailed layer objects we store.  See LayerChildrenSet for the
-    # logic on when that change over happens
-    #
-    # QUESTION: What do we do if we attempt to merge an item that has children?
-    class MergedLayer
-      attr_reader :type
+  # A MergedLayer is a lossy-compression approach to fall back on once we max out
+  # the number of detailed layer objects we store.  See LayerChildrenSet for the
+  # logic on when that change over happens
+  #
+  # QUESTION: What do we do if we attempt to merge an item that has children?
+  class MergedLayer
+    attr_reader :type
 
-      def initialize(type)
-        @type = type
+    def initialize(type)
+      @type = type
 
-        @total_call_time = 0
-        @total_exclusive_time = 0
-        @total_allocations = 0
-        @total_exclusive_allocations = 0
-        @total_layers = 0
-      end
+      @total_call_time = 0
+      @total_exclusive_time = 0
+      @total_allocations = 0
+      @total_exclusive_allocations = 0
+      @total_layers = 0
+    end
 
-      def absorb(layer)
-        @total_layers += 1
+    def absorb(layer)
+      @total_layers += 1
 
-        @total_call_time += layer.total_call_time
-        @total_exclusive_time += layer.total_exclusive_time
+      @total_call_time += layer.total_call_time
+      @total_exclusive_time += layer.total_exclusive_time
 
-        @total_allocations += layer.total_allocations
-        @total_exclusive_allocations += layer.total_exclusive_allocations
-      end
+      @total_allocations += layer.total_allocations
+      @total_exclusive_allocations += layer.total_exclusive_allocations
+    end
 
-      def total_call_time
-        @total_call_time
-      end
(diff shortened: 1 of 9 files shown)


itsderek23:
Thanks @cschneid!

Can you chat w/ @mattscoutapp about any UI implications for handling traces w/merged layers?

cschneid:
Initially, this is all that's required to show `Merged` correctly for AR. I'll make sure that HTTP and such work too.

```
in apm/ on master
› gd lib
diff --git a/lib/slow_tree/node.rb b/lib/slow_tree/node.rb
index f2f18b7..d5aaeb3 100644
--- a/lib/slow_tree/node.rb
+++ b/lib/slow_tree/node.rb
@@ -103,7 +103,12 @@ module SlowTree
             "SQL#Unknown"
           end
         else
-          "SQL#Unknown"
+          # TODO: Add a note next to this line item explaining what merged means
+          if name == "ActiveRecord/Merged"
+            "SQL#Merged"
+          else
+            "SQL#Unknown"
+          end
         end
       else
         MetricFormatter.new(name, app).to_s

```

cschneid:
We'll probably want to add an info bubble next to it that explains why we merged & stopped recording at some point.

lib/scout_apm/merged_layer.rb
```diff
@@ -0,0 +1,124 @@
...
+      end
+
+
+      #######################################################################
+      #  Many methods don't make any sense on a merged layer. Raise errors  #
+      #  aggressively for now to detect mistaken calls                      #
+      #######################################################################
+
+      def add_child
+        raise "Should never call add_child on a merged_layer"
```
dlanderson: Remind me what happens if we raise here?

lib/scout_apm/merged_layer.rb
```diff
@@ -0,0 +1,124 @@
...
+      def subscopable?
+        false
+      end
+
+      def desc
+        nil
+      end
+
+      def backtrace
+        nil
```
dlanderson: What will the user see in the UI? If the backtrace is nil, will clicking a 'SQL#Merged' button do nothing?

cschneid: `nil` is the zero value of backtrace for all layers.  So no biggie if this stays that way.  The view won't render any backtrace button at all.

lib/scout_apm/merged_layer.rb
cschneid: Likely we 500 a request.

Two things though:

* this is a "creation" method on a layer. It is only called by `TrackedRequest`, and  `TrackedRequest` never has a handle to hold onto a `MergedLayer`, so it can't ever call this method.
* calling this function really is undefined. We could swallow it, but that is silently swallowing a real bug.

lib/scout_apm/merged_layer.rb
```diff
@@ -0,0 +1,124 @@
+module ScoutApm
+  module Layers
+    # A MergedLayer is a lossy-compression approach to fall back on once we max out
+    # the number of detailed layer objects we store.  See LayerChildrenSet for the
+    # logic on when that change over happens
+    #
+    # QUESTION: What do we do if we attempt to merge an item that has children?
+    class MergedLayer
```
cschneid: I wonder if this should be named `LimitedLayer` to allow the other (future) optimization of merging layers as we accumulate them ("Oh, another of your N+1 query showed up, we'll count that in our one spot rather than making a whole new layer"), which would take the `MergedLayer` name.

lib/scout_apm/merged_layer.rb
dlanderson: Ok - does that hold true for the rest of the methods below, as well?

dlanderson:
👍  on having something in the UI explaining what 'Foo#Merged' means.

PR #82, opened 2016-11-29
Support extracting the Job name for DelayedJob called via ActiveJob

cschneid:


cschneid:
Will be released in version 2.1.16

PR #83, opened 2016-12-15
Remove `Capacity` sampler

cschneid:
The `Capacity` sampler was an experiment that didn't pan out - trying to read the % of time spent serving requests vs the % of time idle - as a reading of how much headroom your server had.  It didn't give very good readings and was never released.

For similar metrics, the QueueTime metric is a good way to tell if your app server is "keeping up".


itsderek23:
👍

PR #84, opened 2016-12-20
Refactor DevTrace's middleware

cschneid:
* Much cleaner, split into different methods.
* Safer - fallback rescue should handle most situations
* Now works with Grape endpoints that return HTML but not ActionDispatch::Response objects

### DIFF
--- a/lib/scout_apm/instant/middleware.rb
+++ b/lib/scout_apm/instant/middleware.rb
@@ -5,6 +5,11 @@ module ScoutApm
     class Page
       def initialize(html)
         @html = html
+
+        if html.is_a?(Array)
+          @html = html.inject("") { |memo, str| memo + str }
+        end
+
         @to_add_to_head = []
         @to_add_to_body = []
       end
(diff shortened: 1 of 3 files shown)


itsderek23:
@cschneid - The structure seems much cleaner. 👍

@mattaschmann or @dlanderson - mind doing a review w/ @cschneid on the details? Believe this touches every request, so it is an important path.

lib/scout_apm/instant/middleware.rb
```diff
@@ -45,66 +50,185 @@ def initialize(app)
...
+        if trace.nil?
+          logger.debug("DevTrace: in middleware, dev_trace is active, and response has a body, but no trace was found. Path=#{path}; ContentType=#{content_type}")
+          return false
+        end
+
+        true
+      end
+
+      def dev_trace_disabled?
+        ! ScoutApm::Agent.instance.config.value('dev_trace')
```
itsderek23: Curious: is this memo-ized at any level? Wondering since it is touched on every request.

cschneid: This middleware is only included in Rails `development` mode, none of this code should ever run in production.

In dev mode, I'm not too worried, this is checked really early in the sequence here. But we could avoid making this class at all I suppose, by checking this config setting in the Middleware, instead of waiting until we created the class and `.call`ed it.

cschneid:
What's left:

- [ ] Add some tests, do another pass on refactoring - see if the ResponseManipulator can be split
- [ ] Look at Rails 4.2+, which appears to have changed the rack response class
- [ ] Rename from `Instant` -> `DevTrace`, (instant was a development name for this, DevTrace is what we're using)

PR #89, opened 2017-01-10
Use "var" keyword here

etipton:
To prevent a "race condition" that can occur when the `setTimeout` functionality is executed after the [incorrectly] global var gets clobbered by another call to `onload`.

### DIFF
--- a/lib/scout_apm/instant/assets/xmlhttp_instrumentation.html
+++ b/lib/scout_apm/instant/assets/xmlhttp_instrumentation.html
@@ -3,7 +3,7 @@
   (function(){var open=window.XMLHttpRequest.prototype.open;var send=window.XMLHttpRequest.prototype.send;function openReplacement(method,url,async,user,password){this._url=url;return open.apply(this,arguments);}
     function sendReplacement(data){if(this.onload){this._onload=this.onload;}
       this.onload=onLoadReplacement;return send.apply(this,arguments);}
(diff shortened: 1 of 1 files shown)


etipton:
Additionally this line in scout_instant.js needs some cleanup I think:

`payload=JSON.parse(raw)`

should probably be

`var payload = JSON.parse(raw);`

(Use a local variable + add a semicolon)

etipton:
Tangential suggestion:

Can you guys set the `Access-Control-Allow-Origin` header to `*` and use `crossorigin='anonymous'` in the `<script>` tag?

More details here: https://rollbar.com/Rollbar/rollbar.js/items/134/solutions/

I ran into this particular bug in development, but by setting those things, it would allow Rollbar / other js error reporting tools to fully report similar errors if they occur "in the wild" in production.

cschneid:
Thank you for the fix - I'll take a look at it and get it released with the next version.

You're right this JS needs another pass of cleanup.

Regarding the Allow-Origin stuff - I'll pass that to our sysadmin who can get that setup.  But do notice that this instant js is only ever included in development mode.  The code that does the injection to make DevTrace work never gets loaded in production.

etipton:
@cschneid Thanks! I didn't realize that about instant js being dev-only so thought I was fixing a potential prod bug, ha. (We're doing a refactor of how our js files are loaded). Anyway, thanks for taking a look!

etipton:
Thanks again @cschneid!

PR #102, opened 2017-06-20
Add automatic instrumentation of API & Metal controllers

cschneid:
The big gotcha I had to work around here is that API & Base controllers both inherit from Metal. So if I instrument both, the code needs to be sure not to start two Controller layers.  See the specific commit messages for the mechanism I used for that (the dynamic module stuff)

### DIFF
--- a/lib/scout_apm/agent.rb
+++ b/lib/scout_apm/agent.rb
@@ -297,6 +297,7 @@ module ScoutApm
         end
       end
 
+      install_instrument(ScoutApm::Instruments::ActionView)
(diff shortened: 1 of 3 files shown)


itsderek23:
@cschneid - how would this interact with folks that are manually including the instrumentation per our current docs?

cschneid:
@itsderek23 the last commit I pushed renamed some things. The tl;dr of it is: the existing code in our docs will continue to "work" (not crash), but will be a no-op. The automatic instrumentation will take over and do its thing. So existing customers shouldn't have any issues continuing to use the code in their repo, but could remove that include whenever they like.

PR #104, opened 2017-06-26
Abstract recording approach from TrackedRequest

cschneid:
This allows different approaches on how to handle the TrackedRequest
object after the request is finished.

This PR adds a configuration option to allow tracing of tracked requests
in a different thread than the request itself, allowing it to get out of
the critical path of the response.

With async_recording setting:

```
Before:

Start Req --> End Req --> #record! (& process) --> Store --> Response sent

After:

Start Req --> End Req --> Response sent
                      --> #record! (& process) --> Store
```

The goal is that the time spent recording the details of the request is
now not in the way of returning the request.  The same amount of work is
done, just in a background thread instead of the hot-path of the
request.

### DIFF
--- a/lib/scout_apm/agent.rb
+++ b/lib/scout_apm/agent.rb
@@ -10,6 +10,7 @@ module ScoutApm
 
     # Accessors below are for associated classes
     attr_accessor :store
+    attr_reader :recorder
     attr_accessor :layaway
     attr_accessor :config
     attr_accessor :logger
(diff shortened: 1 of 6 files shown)


lib/scout_apm/agent.rb
```diff
@@ -118,6 +123,8 @@ def start(options = {})
       init_logger
       logger.info "Attempting to start Scout Agent [#{ScoutApm::VERSION}] on [#{environment.hostname}]"
 
+      @recorder = create_recorder
```
dlanderson: This is called twice?

cschneid: Once in agent startup, and then once in background worker start. Probably is unnecessary with further change I made to async - since it restarts the thread now if the thread stops.

PR #118, opened 2017-08-18
Db query metrics

dlanderson:
Initial database query metrics. WIP.


lib/scout_apm/db_query_metric_stats.rb
```diff
@@ -0,0 +1,59 @@
...
+      self.min_call_time = other.min_call_time if self.min_call_time.zero? or other.min_call_time < self.min_call_time
+      self.max_call_time = other.max_call_time if other.max_call_time > self.max_call_time
+      self.histogram.combine!(other.histogram)
+      self
+    end
+
+    # To avoid conflicts with different JSON libaries handle JSON ourselves.
+    # Time-based metrics are converted to milliseconds from seconds.
+    def to_json(*a)
+       %Q[{"call_count":#{call_count},"total_call_time":#{call_time*1000},"min_call_time":#{min_call_time*1000},"max_call_time":#{max_call_time*1000},"rows_returned":#{rows_returned}}, "histogram":#{histogram.as_json}]
```
cschneid: Break this into several lines - no need to prematurely compress this string like this, we gzip it.

lib/scout_apm/layer_converters/database_converter.rb
```diff
@@ -0,0 +1,25 @@
...
+        scope = scope_layer
+
+        # TODO: Track requests that never reach a Controller (for example, when
+        # Middleware decides to return rather than passing onward)
+        return {} unless scope
+
+        create_database_metrics
+      end
+
+      def create_database_metrics
```
cschneid: This doesn't create metrics, it returns a list of layers

lib/scout_apm/serializers/payload_serializer_to_json.rb
```diff
@@ -11,6 +11,7 @@ def serialize(metadata, metrics, slow_transactions, jobs, slow_jobs, histograms)
                         :jobs => JobsSerializerToJson.new(jobs).as_json,
                         :slow_jobs => SlowJobsSerializerToJson.new(slow_jobs).as_json,
                         :histograms => HistogramsSerializerToJson.new(histograms).as_json,
+                        :db_query_metrics => DbQuerySerializerToJson.new(db_query_metrics).as_json
```
cschneid: trailing comma plz.

lib/scout_apm/store.rb
```diff
@@ -23,29 +23,36 @@ def current_period
       reporting_periods[current_timestamp]
     end
 
+    def find_period(timestamp = nil)
+      if timestamp
+        @reporting_periods[options[:timestamp]]
```
cschneid: `options` doesn't exist here.

lib/scout_apm/db_query_metric_set.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    def lookup(other)
+      metrics[other.key] ||= other
+    end
+
+    def combine!(other)
+      lookup(other).combine!(other)
+    end
+
+    # Takes an array of ActiveRecord layers, creates new DbQueryMetricStats and combines them into this Set
```
cschneid: Not sure I like this class knowing so much about how we store layers. Elsewhere we stop caring that the data was in layers after the converter step, but in the db case it goes all the way through the Store, into this class.

Can we create those DbQueryMetricStats a lot earlier, and have this class take those as its input?

joelr:
😬

PR #119, opened 2017-08-25
Also find queue of a resque job from klass.queue

cschneid:
Also add a fallback. We shouldn't ever allow queue to be nil, even
though I don't think resque will run a job if both @queue and #queue are
missing. Always have a fallback of "unknown", so we can report smething.

This oversight of queue name was manifesting as a call to nil.downcase
when we go to process the JobRecord.


cschneid:
Resque docs on what is valid to #dequeue http://www.rubydoc.info/github/resque/resque/Resque#dequeue-instance_method

dlanderson:
👍

PR #120, opened 2017-08-29
Fix Rails 5+ SQL parameter logging

qrush:
We were seeing no SQL parameters logging to our normal Rails logs and this was why!

I'd like to encourage Scout to please use ActiveRecord's instrumentation instead of overriding methods. This could just be a subscription to `sql.active_record` now instead of messing with alias method chains.

https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L542-L543

### DIFF
--- a/lib/scout_apm/instruments/active_record.rb
+++ b/lib/scout_apm/instruments/active_record.rb
@@ -122,7 +122,7 @@ module ScoutApm
             current_layer.desc = desc
           end
 
-          log_without_scout_instruments(sql, name, &block)
+          log_without_scout_instruments(*args, &block)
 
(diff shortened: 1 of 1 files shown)


cschneid:
Thank you for the PR. Sorry it caused you some problems.

I will look into using the AR subscription instead, but the typical gotcha I have run into trying to move to doing things the "right" way is trying to support old versions of Ruby and Rails, and also be sure that I get the actual data I need (cached queries vs not in this case).

I'll give it a try. Thanks again for the PR.

qrush:
Thanks. Would appreciate a new gem release too so we can get off my fork. 🙏

cschneid:
@qrush - released in 2.1.31. Thanks again!

PR #138, opened 2017-12-19
Be more generous with max sql length we're willing to scrub

cschneid:
SQL Scrubbing is only used when we capture traces, which only happens a
handful of times per minute, so performance isn't as big a deal now as
when we initially introduced this limit. Back then, we were running sql
scrubbing eagerly on nearly every SQL statement


itsderek23:
Thanks @cschneid! Looks good!

Issue #149, opened 2018-01-16
APM 2.4.x passenger integration called removed method `shutdown`


PikachuEXE:
https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/server_integrations/passenger.rb#L32

While the `ExitHandler` seems to be updated
https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/agent/exit_handler.rb

Need PR?

PR #150, opened 2018-01-16
Fix passenger using removed method

PikachuEXE:
Fix https://github.com/scoutapp/scout_apm_ruby/issues/149

Test locally with passenger 5.1.3 open source version


cschneid:
Thank you for this PR! Just merged & released in 2.4.2.

PR #156, opened 2018-01-25
Ensure instruments are installed correctly

cschneid:
There was logic to avoid installing instruments when an app server wasn't
detected, and also separate logic to start the agent when a web request
came through. The interaction was that it would start the agent, but not
actually install instruments. So an app would report memory & cpu usage,
but not any throughput or traces.

This ensures that instruments are installed more often (when monitor=true
at least)

### DIFF
--- a/lib/scout_apm/agent.rb
+++ b/lib/scout_apm/agent.rb
@@ -35,15 +35,14 @@ module ScoutApm
     def install(force=false)
       context.config = ScoutApm::Config.with_file(context, context.config.value("config_file"))
 
-      context.logger.info "Scout Agent [#{ScoutApm::VERSION}] Initialized"
+      logger.info "Scout Agent [#{ScoutApm::VERSION}] Initialized"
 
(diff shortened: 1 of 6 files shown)


lib/scout_apm/agent.rb
```diff
@@ -69,6 +68,8 @@ def start(opts={})
 
       install unless context.installed?
 
+      instrument_manager.install! # noop if it was installed earlier
```
dlanderson: Does this need an `if should_load_instruments?` check?

cschneid: no, since all that would do is check `monitor`, which we did at the top of this method anyway.

dlanderson: What if `should_install_instruments?` changes? Leave that decision up to `should_install_instruments?` instead of assuming it will be the same as the `monitor` check on L62. That also makes it consistent with how it's used on L40

cschneid: Added in 3950108 after I merged

Issue #145, opened 2018-01-04
layer_finder.job is nil in JobConverter which gives undefined method `legacy_metric_name`


jpslav:
We just upgraded from 3.0.0.pre13 to 3.0.0.pre16 in four different repositories.  For some reason, one of the repos is giving us a bunch of `NoMethodError: undefined method `legacy_metric_name' for nil:NilClass` errors in our specs (https://travis-ci.org/openstax/tutor-server/jobs/325192670).  

With debugger found this is happening at https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/layer_converters/job_converter.rb#L21 :

```ruby
@meta_options = {:scope => layer_finder.job.legacy_metric_name}
```

`layer_finder.job` is `nil`.  The `request.job?` check several lines up did return `true`.

Any ideas?

cschneid:
It's odd we'd be running in test mode at all. Check that `monitor: false` is in the test part of scout_apm.yml.

Does the list of failing tests stay stable between runs?

I wonder if it's related to the thread-local "current-transaction" that is running. If it gets in a confused state, I could see it answering 'yes' to being a job, but not actually having the data inside that is required. You can clear any request by doing: `Thread.current[:scout_request] = nil`.

jpslav:
@cschneid There isn't even a `scout_apm.yml` file in our repository -- we only write that config file to our deployed production servers.  (here's the repo https://github.com/openstax/tutor-server/tree/scout-apm-update)

I'm not sure if the failure list is consistent, but I am able to replicate the error over and over in any given test.  

I reverted to `3.0.0.pre13` and the spec passed.  I upgraded one step at a time and saw the failure when moving to `3.0.0.pre14`.

When I am in the failing spec, I get:

```ruby
(byebug) ScoutApm::Agent.instance.context.config.value("monitor")
false
```

Inside the failing spec, `Thread.current[:scout_request]` is definitely not nil, and inside the `TrackedRequest` object used in the `JobConverter`, I get 

```ruby
(byebug) @agent_context.config.value("monitor")
false
```

cschneid:
@jpslav your debugging is amazing. If that repo reproduces the error, I'll fix it, and let you know. I'm sorry for the hassle.

jpslav:
[...]

PR #159, opened 2018-02-01
Fix recording backtraces

cschneid:
When #145 was fixed, we changed the definition of the `web?` and `job?`
methods in TrackedRequest to look at the real stack of layers, rather
than using marker booleans set by the ActionController and background
job instrumentations.

Because of that change, and caching of the values, an optimization
check in `capture_backtrace?` was always returning false, skipping
backtrace collection.

The fix is to untangle the logical `web?` call into two distinct items.
One is for recording, and is still called `web?` (and `job?`). It looks
at the layers.

Then adding a new flag, set while a request is running, that marks it as
a `real_request!`. That then is a signal to future instrumentation that
its worthwhile to enable all the various captures we do, and that we're
not wasting time.

The initial reason for this optimization was that Sidekiq is constantly
churning TrackedRequest objects, every time it checks Redis (and other
bookkeeping tasks). We found that recording backtraces was rather
expensive in that case, and they were just being collected, then thrown
away.

### DIFF
--- a/lib/scout_apm/tracked_request.rb
+++ b/lib/scout_apm/tracked_request.rb
@@ -38,6 +38,13 @@ module ScoutApm
     # An object that responds to `record!(TrackedRequest)` to store this tracked request
     attr_reader :recorder
 
+    # When we see these layers, it means a real request is going through the
+    # system. We toggle a flag to turn on some slightly more expensive
+    # instrumentation (backtrace collection and the like) that would be too
+    # expensive in situations where the framework is constantly churning. We
+    # see that on Sidekiq.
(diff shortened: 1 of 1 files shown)


cschneid:
This issue includes 2.4.0 through 2.4.6, and the fix will be released in 2.4.7.

lib/scout_apm/tracked_request.rb
```diff
@@ -105,6 +116,15 @@ def stop_layer
       end
     end
 
+    def real_request!
+      @real_request = true
```
dlanderson: Perhaps default this to `false` in `initialize` just to be explicit

PR #161, opened 2018-02-16
Extracts ScoutApm::Tracer.instrument

### LINKED ISSUE #160: Create a direct way to call ScoutApm::Tracer.instrument with a block
Right now, the code relies on being included as a module into the tracing class. We should have a way to instrument by calling into `instrument` directly without including anything.

```ruby
def foo
  # prefix code

  ScoutApm::Tracer.instrument("Custom", "FooWork") do
    # Code I really care about tracing
  end

  # postfix code
end
```

alindeman:
While this is a pretty simple extraction, I used it as an opportunity to try to write a test or few and also poke around other bits of the codebase. I'm also interested in feedback about whether these tests are valuable.

Resolves #160

### DIFF
--- a/lib/scout_apm/tracer.rb
+++ b/lib/scout_apm/tracer.rb
@@ -12,31 +12,36 @@ module ScoutApm
       klass.extend ClassMethods
     end
 
+    # Type: the Layer type - "View" or similar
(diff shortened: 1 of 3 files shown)


cschneid:
The tests are great, I added one more to explicitly check that we can instrument just a section of a method.

FYI - we have travis setup at https://travis-ci.org/scoutapp/scout_apm_ruby

I'll merge this as soon as travis runs.

Issue #140, opened 2017-12-21
API for manually setting name of an endpoint/job


cschneid:
We've run into the use case of a central Controller or Job that then determines the actual work to do based on parameters (think `GraphQL` controller, or a `EventProcessor` background job). Since they're acting as user-driven routers, it's more useful to name the work done based on the result of that routing, not just the router itself.

Right now, we have a rough way to rename a transaction:

```ruby
req = ScoutApm::RequestManager.lookup
req.current_layer.name = "Override"
```

But that has limitations, and works only if the current_layer is exactly the `Controller` or `Job` layer.

A better api could be created, making it easier to use, and also hard to misuse (by checking the arity of that name and other similar gotchas).

alindeman:
What is your sense of what the API should look like? A new method on the `TrackedRequest`, like it looks like [we've recommended before](https://github.com/scoutapp/scout_apm_ruby/blob/master/CHANGELOG.markdown#2114), perhaps?

cschneid:
I think an API that works more like how `Context` works would be best.

```
ScoutApm::Transaction.ignore!
ScoutApm::Transaction.rename("User Lookup")
```

(the word Transaction since then it works for both web requests and background jobs).

Also note that `rename` should be smarter than simply changing current layer. It should go find the Controller or Job layer and rename that. That way we don't accidentally rename a different, nested, layer when called. Makes it less fidlly for users.

cschneid:
Closed by #162

PR #162, opened 2018-02-19
Introduces ScoutApm::Transaction to manipulate requests

alindeman:
Specifically, two new methods:
* `ignore!` as a more convenient way to call `RequestManager.lookup.ignore_request!`
* `rename` as a way to find the most recent Controller or Job layer and rename it

Open questions:
- [ ] Is this logic for finding the last Controller or Job layer correct?
- [ ] What happens if there is no Controller or Job layer? Should this `raise` or do something other than do nothing?
- [x] Run this code against a sample Rails app

Relates #140


lib/scout_apm/tracked_request.rb
```diff
@@ -267,6 +271,14 @@ def record!
       # Bail out early if the user asked us to ignore this uri
       return if @agent_context.ignored_uris.ignore?(annotations[:uri])
 
+      # Apply name override if needed
+      unless name_override.nil?
+        scope_layer = layer_finder.scope
+        unless scope_layer.nil?
+          scope_layer.name = name_override
+        end
+      end
```
alindeman: @cschneid As I was poking around a bit more, I found this `scope` layer concept. Is this a reasonable way to achieve what we want here? Or does the rename need to be a `LayerConverter`?

test/unit/tracked_request_test.rb
```diff
@@ -10,19 +10,6 @@ def test_marshal_dump_load
...
-    context = ScoutApm::AgentContext.new
-    tr = ScoutApm::TrackedRequest.new(context, faux_store)
-    assert_equal faux_store, tr.instance_variable_get("@store")
-
-    tr.prepare_to_dump!
-    assert_nil tr.instance_variable_get("@store")
-
-    tr.restore_store
-    assert_equal context.store, tr.instance_variable_get("@store")
-  end
```
alindeman: This test file wasn't running because it was named `test_...` instead of `..._test.rb`. When I fixed that, this test failed because it's calling a method that no longer exists.

lib/scout_apm/tracked_request.rb
cschneid: `if name_override` and `if scope_layer` - prefer positive conditions.

This is right though, pull it out into a method like `apply_name_override()` or something?

alindeman:
@cschneid This is ready for another look. Does the public API make sense? Is the implementation solid?

I did have a chance to try this with a local dev app:

```ruby
class PostsController < ApplicationController
  def index                              
    ScoutApm::Transaction.rename("posts/foobar")                                   
    @posts = Post.all                    
  end
end
```

And it showed up correctly, as far as I can tell:

screenshot 2018-02-27 at 22 09 36

cschneid:
Looks great - thank you.

itsderek23:
When this is released, merge in https://github.com/scoutapp/slate_apm_help/pull/17. This adds doc info.

PR #164, opened 2018-02-19
Adds Rubies (including 1.8.7) to Travis build matrix

alindeman:
Developing for 1.8.7 can be tough, and so I think it's important that 1.8.7 be in the build matrix.

There were some actual issues, at least with tests, in the current codebase. These should be fixed, or if I felt like they couldn't be fixed but weren't providing good value on 1.8.7, they are skipped.

Some developer tooling gems just refuse to install and have been conditionally removed on 1.8.7.

Outstanding questions:
- [ ] Is the order that children are walked within a level important? For instance, if A has children B and C, does it matter if the walker walks B or C first? If so, that's currently broken on 1.8.7 because 1.8.7 doesn't keep hash keys in insertion order.
- [ ] Create issue for adding `rubocop` to the Travis build


cschneid:
👍  Great changeset.

Re: ordering - The reporting we do doesn't depend on children ordering. But I agree that the test you call out for layer ordering would get rather complicated to have it support 1.8. We can leave it as skipped.

Rubocop in travis sounds good, go ahead and add that.

alindeman:
@cschneid Thanks for the context. I've added a few more commits. What do you think?

PR #168, opened 2018-03-01
Pins rake to 10.5 on Ruby 1.8.7

alindeman:
Not sure why this didn't show up as an issue until now, but this should fix the Travis build failure on Ruby 1.8.7.


cschneid:
Thanks! I saw this yesterday too, and have no idea why it only cropped up now.

PR #169, opened 2018-03-01
Preconditions have configurable severity

### LINKED ISSUE #130: Too much "info" logs
Hello,

My server logs are full of "info" logs written by scout, which should be - I think - written as "debug" logs.
I don't really care what scout is "instrumenting"  or "attempting".

The only logs that can be "info" are the last 2 lines or the init :

INFO : Scout Agent [2.1.32] Initialized
INFO : Sending Application Startup Info - App Server: webrick, Framework: Rails, Framework Version: 5.0.6, Database Engine: postgres

Thank you,

Damien

alindeman:
I guessed a bit at each `severity` here. I'm open to changing them based on feedback.

Resolves #130

### DIFF
--- a/lib/scout_apm/agent/preconditions.rb
+++ b/lib/scout_apm/agent/preconditions.rb
@@ -4,20 +4,24 @@ module ScoutApm
       # The preconditions here must be a 2 element hash, with :message and :check.
       # message: Proc that takes the environment, and returns a string
       # check: Proc that takes an AgentContext and returns true if precondition was met, if false, we shouldn't start.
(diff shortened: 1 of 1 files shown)


alindeman:
@dlanderson WDYT?

cschneid:
Great - Those levels look good. Info for stuff that the user could reasonably expect to leave as-is (running in rails console, or monitor false), and warn for things that indicate that the agent isn't installed right (app name, and the legacy scout agent).

dlanderson:
👍

PR #171, opened 2018-03-13
Change stale layaway regex, add test

dlanderson:
We had a customer with a `data_file` set to `/srv/appname/shared/tmp/scout_apm/`. The `scout_apm` part of the path was causing the `Layaway#timestamp_from_filename` used by `Layaway#delete_stale_files(older_than)` to pick up the wrong file name:

```
DEBUG : Deleting stale layaway files with timestamps: ["apm/scout_201803121007", "apm/scout_201803102318"
```
This PR changes the `match` regex used by `Layaway#timestamp_from_filename` which should fix the issue. This PR also adds a test for this case but for some reason the test is not running as expected. @cschneid or @alindeman mind taking a look at the test?

### DIFF
--- a/lib/scout_apm/layaway.rb
+++ b/lib/scout_apm/layaway.rb
@@ -188,7 +188,7 @@ module ScoutApm
     end
 
     def timestamp_from_filename(filename)
-      match = filename.match(%r{scout_(.*)_.*\.data})
+      match = filename.match(%r{scout_(\d+)_\d+\.data})
       if match
         match[1]
       else
--- a/lib/scout_apm/layaway.rb
+++ b/lib/scout_apm/layaway.rb
@@ -188,7 +188,7 @@ module ScoutApm
     end
 
     def timestamp_from_filename(filename)
-      match = filename.match(%r{scout_(\d+)_\d+\.data})
+      match = filename.match(%r{scout_(\d+)_\d+\.data\z})
       if match
         match[1]
       else
--- a/test/unit/layaway_test.rb
+++ b/test/unit/layaway_test.rb
@@ -58,23 +58,27 @@ class LayawayTest < Minitest::Test
     context = ScoutApm::AgentContext.new().tap{|c| c.config = config; c.environment = env }
     layaway = ScoutApm::Layaway.new(context)
 
-    not_stale_time_integer = Time.now.strftime(ScoutApm::Layaway::TIME_FORMAT).to_i
-    stale_time_integer = not_stale_time_integer - (ScoutApm::Layaway::STALE_AGE + 2)
 
-    not_stale_file_names = [File.join(data_dir, "scout_#{not_stale_time_integer}_1.data"),
-                            File.join(data_dir, "scout_#{not_stale_time_integer}_20.data")]
-    stale_file_names = [File.join(data_dir, "scout_#{stale_time_integer}_1.data"),
-                        File.join(data_dir, "scout_#{stale_time_integer}_20.data")]
+    not_stale_time = Time.now
+    not_stale_time_formatted = not_stale_time.strftime(ScoutApm::Layaway::TIME_FORMAT)
+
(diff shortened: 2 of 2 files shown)


test/unit/layaway_test.rb
```diff
@@ -46,4 +46,35 @@ def test_layaway_file_limit_prevents_new_writes
...
+    all_file_names = not_stale_file_names + stale_file_names
+
+    (all_file_names).each do |filename|
+        File.new(filename, 'w')
+    end
+
+    assert_equal Pathname.new("/tmp/scout_apm_test/shared/scout_apm"), ScoutApm::Layaway.new(context).directory
+    assert_equal all_file_names.sort, Dir.glob("#{data_dir}/*data").sort
+
+    layaway.delete_stale_files(not_stale_time_integer - ScoutApm::Layaway::STALE_AGE)
```
alindeman: This suffers from the same time subtraction issue. And also, `Layaway#delete_stale_files` expects a timestamp as an instance of `Time`. What ends up happening is `delete_stale_files` raises an exception, but [it's then logged and discarded](https://github.com/scoutapp/scout_apm_ruby/blob/97900151e9dbfe26132ab881e34445943679aa26/lib/scout_apm/layaway.rb#L147):

```
[1] pry(#<ScoutApm::Layaway>)> e
=> #<NoMethodError: undefined method `strftime' for 201803140459:Integer>
```

When we have a reference to the original `Time` in a variable like `not_stale_time`, it can be used here.

lib/scout_apm/layaway.rb
```diff
@@ -188,7 +188,7 @@ def format_pid(pid)
     end
 
     def timestamp_from_filename(filename)
-      match = filename.match(%r{scout_(.*)_.*\.data})
+      match = filename.match(%r{scout_(\d+)_\d+\.data})
```
alindeman: What do you think of also adding a `\z` to the end of this regex, to make it clear the filename must be at the end of the string?

test/unit/layaway_test.rb
```diff
@@ -46,4 +46,35 @@ def test_layaway_file_limit_prevents_new_writes
...
+    # Clean out files
+    FileUtils.safe_unlink(Dir.glob("#{data_dir}/scout_*_*.data"))
+
+    config = make_fake_config({'data_file' => data_dir})
+    env = make_fake_environment(:root => '/tmp/scout_apm_test')
+    context = ScoutApm::AgentContext.new().tap{|c| c.config = config; c.environment = env }
+    layaway = ScoutApm::Layaway.new(context)
+
+    not_stale_time_integer = Time.now.strftime(ScoutApm::Layaway::TIME_FORMAT).to_i
+    stale_time_integer = not_stale_time_integer - (ScoutApm::Layaway::STALE_AGE + 2)
```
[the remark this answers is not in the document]
dlanderson: Good suggestion. Switching between the `Time` objects and time formated strings in `layaway.rb` tripped me up.

test/unit/layaway_test.rb
```diff
@@ -46,4 +46,35 @@ def test_layaway_file_limit_prevents_new_writes
...
+
+    not_stale_file_names = [File.join(data_dir, "scout_#{not_stale_time_integer}_1.data"),
+                            File.join(data_dir, "scout_#{not_stale_time_integer}_20.data")]
+    stale_file_names = [File.join(data_dir, "scout_#{stale_time_integer}_1.data"),
+                        File.join(data_dir, "scout_#{stale_time_integer}_20.data")]
+    all_file_names = not_stale_file_names + stale_file_names
+
+    (all_file_names).each do |filename|
+        File.new(filename, 'w')
```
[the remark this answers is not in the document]
dlanderson: Been in Python land for a while - fixed to 2 spaces. I added `.close` on the file handle, but touch would also work.

test/unit/layaway_test.rb
dlanderson: Separated `Time` objects and clarified in 05d4f81

dlanderson:
Thanks for the investigation and review @alindeman ! Test passes. Resolves https://github.com/scoutapp/scout_apm_ruby/issues/172

PR #178, opened 2018-04-03
[Work in Progress] Supports Method#prepend for Tracer

### LINKED ISSUE #114: Should not use alias_method
It is risky if any other gems use `prepend` against the same method.
https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/tracer.rb#L63-L64

see also:
https://speakerdeck.com/a_matsuda/ruby-2-in-ruby-on-rails (page 37-43)
> It's dangerous to use both AMC and `prepend` against a single method
> The result is not only unpredictable
> Sometimes critical ☠️

alindeman:
This still needs some work and further testing, but I felt like it would be good to go ahead and get it up for review.

If supported (Ruby 2.0 and above), methods instrumented using `Tracer` will used an anonymous module `prepend`ed to the class. This means the method will appear first in the ancestor chain, and can call `super` instead of `foo_without_scout_instrument`. We want this because `Module#prepend` is usually incompatible with `alias_method_chain`, and can result in infinite loops. It should help partially res olve #114.

I want to separate this work from work on specific instruments (such as ActiveRecord) because those modifications will be less generic and need specific testing.

Some other open questions:
* Should there be a way to force `prepend` off with a configuration option? I think that's the safest way to roll this out, because users might be using _other_ gems that do not support `prepend`, but I'm not sure the best way to get access to Scout's config in `Tracer` other than calling `ScoutApm::Tracer.agent.instance.context.config`. How do folks feel about that?
* Does it fully res olve #114's specific issue, when using `ActionView` instrumentation and `jb`? I need to create a test app and try it.

### DIFF
--- a/lib/scout_apm/environment.rb
+++ b/lib/scout_apm/environment.rb
@@ -165,15 +165,18 @@ module ScoutApm
     end
 
     def ruby_19?
-      @ruby_19 ||= defined?(RUBY_ENGINE) && RUBY_ENGINE == "ruby" && RUBY_VERSION.match(/^1\.9/)
+      return @ruby_19 if defined?(@ruby_19)
+      @ruby_19 = defined?(RUBY_ENGINE) && RUBY_ENGINE == "ruby" && RUBY_VERSION.match(/^1\.9/)
     end
 
     def ruby_187?
-      @ruby_187 ||= defined?(RUBY_VERSION) && RUBY_VERSION.match(/^1\.8\.7/)
+      return @ruby_187 if defined?(@ruby_187)
+      @ruby_187 = defined?(RUBY_VERSION) && RUBY_VERSION.match(/^1\.8\.7/)
     end
 
     def ruby_2?
-      @ruby_2 ||= defined?(RUBY_VERSION) && RUBY_VERSION.match(/^2/)
+      return @ruby_2 if defined?(@ruby_2)
+      @ruby_2 = defined?(RUBY_VERSION) && RUBY_VERSION.match(/^2/)
     end
 
     # Returns a string representation of the OS (ex: darwin, linux)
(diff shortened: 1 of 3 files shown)


lib/scout_apm/tracer.rb
```diff
@@ -120,28 +129,54 @@ def _instrumentable?(method_name)
...
+      # Returns a module that is prepended to the class where methods
+      # containing instrumentation code can be added. Methods on this module
+      # should must call `super` to invoke the original method.
+      #
+      # Only used if Module#prepend is supported (Ruby >= 2.0.0)
+      def _scout_instrumentation_module
+        return @_scout_instrumentation_module if defined?(@_scout_instrumentation_module)
+
+        @_scout_instrumentation_module = Module.new
+        prepend(@_scout_instrumentation_module)
```
mjobin-mdsol, line 176: Careful `Module#prepend` is public from 2.1, private in 2.0

lib/scout_apm/environment.rb
```diff
@@ -165,15 +165,23 @@ def jruby?
...
 
     def ruby_2?
-      @ruby_2 ||= defined?(RUBY_VERSION) && RUBY_VERSION.match(/^2/)
+      return @ruby_2 if defined?(@ruby_2)
+      @ruby_2 = defined?(RUBY_VERSION) && RUBY_VERSION.match(/^2/)
+    end
+
+    # Returns true if this Ruby version supports Module#prepend.
+    def supports_module_prepend?
+      ruby_2?
```
mjobin-mdsol, line 184: scout now only supports 2.1+
so most of this PR can be trashed

cschneid:
Closing in favor of using prepend by default, and a configuration option to fall back to alias_method later. Similar work will be done in `Tracer.instrument` to allow adhoc custom instrumentation to make the same choice.

PR #180, opened 2018-04-06
[Work In Progress] Instruments #apply_join_dependency in Rails 5.2 and above

alindeman:
In Rails 5.2, `ActiveRecord::FinderMethods#find_with_associations` was removed and its functionality folded into `#apply_join_dependency`. See [rails/rails#412db710](https://github.com/rails/rails/commit/412db710dfa6ed84654068576b1841966d7f89b2#diff-1cabaf53664d448a5cbfcf396cd842b1).

This is the most straightforward 'fix', but it does require a bit more investigation:
* What was the original intention of starting a layer in `find_with_associations`? Does instrumenting `apply_join_dependency` solve the same thing?
* `apply_join_dependency` is called in more places than `find_with_associations` was, including places like `Relation#to_sql` where the query may not be executed at all. Is that harmful in any way to the instrumentation?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ rvm:
   - "2.4"
   - "2.5"
 cache: bundler
+before_install:
+  - gem update --system
+  - gem install bundler
 jobs:
   include:
     - script: bundle exec rake test
--- a/lib/scout_apm/instruments/active_record.rb
+++ b/lib/scout_apm/instruments/active_record.rb
@@ -208,8 +208,9 @@ module ScoutApm
     # Entry-point of instruments.
     #
     # We instrument both ActiveRecord::Querying#find_by_sql and
-    # ActiveRecord::FinderMethods#find_with_associations.  These are early in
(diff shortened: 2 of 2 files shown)


cschneid:
`find_by_sql` and `find_by_associations` were the two entry-point functions I found when instrumenting ActiveRecord. They were the logical entry point to a single query. Then deeper in AR, the `log` method is the running of a single SQL statement.

By starting a layer higher up the chain, it allows us to trace a single `User.where().include(:posts)` logical query, even if that ends up executing a number of SQL statements.

The `ignorable` logic is because this is before the statement cache gets checked, and we don't record SQL calls unless they go out to the real db. So we need to mark calls at this layer as provisional, and dump them if they never result in a real call.

It's quite possible that we need to re-evaluate how Rails 5.x is doing database calls and find a new point to instrument. The vast majority of SQL calls appear to go through the other `find_by_sql` instrument.

alindeman:
With some more time, I think we can do better than this. Will submit a fresh PR after I have something ready.

Issue #179, opened 2018-04-05
Rails 5.2.0.rc2 causes ActiveRecord instrumentation exception


indirect:
In my Rails 5.2.0.rc2 app, with scout_apm 2.4.9, running any process that enables Scout monitoring results in this log line:

```
[Scout] [04/05/18 08:48:19 +0000 run.6514 (679)] WARN : ActiveRecord instrumentation exception: undefined method `find_with_associations' for module `ActiveRecord::FinderMethods'
```

cschneid:
Yes, we need to check everything against Rails 5.2, which I think is being released imminently.

cschneid:
This was fixed in #182

PR #182, opened 2018-04-09
Instruments early via exec_queries in Rails >= 3.2

alindeman:
Rails 5.2 factored away the `find_with_assocations` method, so a replacement needs to be found for starting instrumentation without generating a lot of false noise.

Rails 3.2 introduced a private method `exec_queries` to `ActiveRecord::Relation` that has survived through to Rails 5.2 and seems relatively stable at this point. It is a good candidate to achieve the same instrumentation goals because it is just one call earlier in the chain of the calls invoked when a query (or group of queries) is being invoked.

In Rails 3.2 through Rails 5.2, the possible call chains relevant to instrumentation look like:
* `exec_queries` -> (if eager loading) _uninstrumented methods_ -> (if cache miss) `log`
* `exec_queries` -> (if not eager loading) `find_by_sql` -> (if cache miss) `log`

Rails 3.2 through Rails 5.1 call `find_with_assocations` if eager loading but it no longer needs to be instrumented because the layer has been started earlier in `exec_queries`.

In the not eager loading case, both `exec_queries` and `find_by_sql` are instrumented but the request is intentionally marked `ignore_children` to avoid starting duplicate layers. The instrumentation in `find_by_sql` is kept because it is possible for client code to directly call `Model.find_by_sql("...")` and that should be instrumented as well.

In previous versions of Rails, the original implementation stays as-is, which looks like:
* (if eager loading) `find_with_associations` -> (if cache miss) `log`
* (if not eager loading) `find_by_sql` -> (if cache miss) `log`

### DIFF
--- a/lib/scout_apm/instruments/active_record.rb
+++ b/lib/scout_apm/instruments/active_record.rb
@@ -112,9 +112,9 @@ module ScoutApm
           end
         end
 
-        rails_3_2_or_above = defined?(::ActiveRecord::VERSION::MAJOR) && \
-          defined?(::ActiveRecord::VERSION::MINOR) && \
-          (::ActiveRecord::VERSION::MAJOR.to_i > 3 || \
+        rails_3_2_or_above = defined?(::ActiveRecord::VERSION::MAJOR) &&
+          defined?(::ActiveRecord::VERSION::MINOR) &&
+          (::ActiveRecord::VERSION::MAJOR.to_i > 3 ||
            (::ActiveRecord::VERSION::MAJOR.to_i == 3 && ::ActiveRecord::VERSION::MINOR.to_i >= 2))
         if rails_3_2_or_above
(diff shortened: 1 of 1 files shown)


cschneid:
Note: the motivation for this change is issue #179.

lib/scout_apm/instruments/active_record.rb
```diff
@@ -112,10 +112,22 @@ def add_instruments
           end
         end
 
-        if Utils::KlassHelper.defined?("ActiveRecord::FinderMethods")
-          ::ActiveRecord::FinderMethods.module_eval do
-            include ::ScoutApm::Tracer
-            include ::ScoutApm::Instruments::ActiveRecordFinderMethodsInstruments
+        rails_3_2_or_above = defined?(::ActiveRecord::VERSION::MAJOR) && \
```
cschneid: The line-continuation backslash isn't needed I don't think?

PR #183, opened 2018-04-09
Extensions: Transaction & Periodic Callbacks

itsderek23:
This adds the ability to fan out data collected by the agent to other services via extensions. 

There are two types of callbacks:

* Transaction callbacks - execute after a transaction completes (a web request or background job)
* Periodic callbacks - execute when data is reported to Scout every minute

There are 2 implementations of this:

* [scout_signalfx_ruby](https://github.com/scoutapp/scout_signalfx_ruby) - periodic callback
* [scout_dogstatsd_ruby](https://github.com/scoutapp/scout_dogstatsd_ruby) - transaction callback


itsderek23:
Rollout steps:

- [ ] Create a `scoutprof_extensions` branch and test in production
- [ ] If ok, update changelog and release 2.4.10.pre. Once tested, release 3.0.0.pre22.

lib/scout_apm/agent_context.rb
```diff
@@ -9,6 +12,7 @@ class AgentContext
     def initialize()
       @logger = LoggerFactory.build_minimal_logger
       @process_start_time = Time.now
+      @extensions = ScoutApm::Extensions::Config.new
```
cschneid: If you pass `self` into `initialize`, you can then reference that local variable later on, rather than reaching back out to the global `AgentContext`

itsderek23: Currently the interface for adding and running extensions is via class methods, not instance methods. With the current design, it feels like it'd be a little confusing to set a class var in `initialize`.

lib/scout_apm/extensions/config.rb
```diff
@@ -0,0 +1,75 @@
...
+      # These run in a background thread so external HTTP calls are OK.
+      # +callback+ must be an object that responds to a +call(reporting_period, metadata)+ method.
+      def self.add_periodic_callback(callback)
+        agent_context.extensions.periodic_callbacks << callback
+      end
+
+      # Runs each reporting period callback. 
+      # Each callback runs inside a begin/rescue block so a broken callback doesn't prevent other
+      # callbacks from executing or reporting data from being sent. 
+      def self.run_periodic_callbacks(reporting_period, metadata)
```
cschneid: This should be an instance method. The class method version can exist, but just to look up the instance from the global AgentContext, and call into it.

lib/scout_apm/extensions/transaction_callback_payload.rb
```diff
@@ -0,0 +1,73 @@
...
+        elsif converter_results[:job]
+          stat = converter_results[:job].metric_set.metrics[ScoutApm::MetricMeta.new("Latency/all", :scope => transaction_name)]
+          stat ? stat.total_call_time*1000 : nil
+        else
+          nil
+        end
+      end
+
+      def hostname
+        ScoutApm::Agent.instance.context.environment.hostname
```
cschneid: The context should get passed in here, rather than reaching out to the global.

itsderek23:
@cschneid - 9f1ca8750d753c36346f8fb034afe2ae2276c564 switches to using instance methods to run extensions.

itsderek23:
This runs fine for me locally w/ the APM app.

PR #191, opened 2018-05-04
Calculates queue latency correctly for scheduled jobs

### LINKED ISSUE #189: Delayed Job Queue Time should account for jobs set to run in the future
Right now, we calculate time in queue as `Time.now - job.created_at`, but if the job was created to run later, that number is incorrect.  More likely, we should calculate it as `Time.now - job.run_at`.

https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/background_job_integrations/delayed_job.rb#L56

alindeman:
If a DelayedJob job is queued for future execution (`Job.delay.deliver(run_at: 5.minutes.from_now)` or via ActiveJob `Job.set(wait: 5.minutes).perform_later`), the queue latency should be calculated from the time the job was expected to run, instead of when it was created.

DelayedJob sets `run_at` to the current time if no delay is specified, so this logic should still work correctly for jobs scheduled for immediate execution.

Fixes #189

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Calculates DelayedJob queue latency correctly when jobs are scheduled to run in the future
+
 # 2.4.11
 
 * Adds transaction + periodic reporting callback extension support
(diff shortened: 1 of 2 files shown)


alindeman:
🤔 I wonder if there are cases when jobs would be scheduled in the past.

```ruby
# some_calculated_timestamp is calculated as being in the past somehow
Job.set(wait_until: some_calculated_timestamp).perform_later
```

DelayedJob will still execute this job, but it'll appear to have significant latency even when there is none in reality.

It's not significantly more difficult to use `[job.created_at, job.run_at].max` to use the most recent of the two. I'll switch it to that instead.

cschneid:
Good thinking 👍

PR #198, opened 2018-06-01
Experimental - DO NOT MERGE! Auto-instrumentation of Controller Time

itsderek23:
Install instructions:

* add `rubocop` to the Gemfile
* add `require 'scout_apm/auto_instrument'` to `config/boot.rb`


itsderek23:
No rush, but I tried adding this to an app, but I'm not seeing any auto-instrumentation. 

Relevant diffs below:

```
diff --git a/Gemfile b/Gemfile
index f467c15..15bf591 100644
--- a/Gemfile
+++ b/Gemfile
@@ -31,9 +31,13 @@
-gem 'scout_apm', git: 'https://github.com/scoutapp/scout_apm_ruby.git', branch: 'scoutprof_statsd_context'
+
+gem 'rubocop' # for auto instrument
+gem 'scout_apm', git: 'https://github.com/scoutapp/scout_apm_ruby.git', branch: 'auto_instruments'
+# gem 'scout_apm', git: 'https://github.com/scoutapp/scout_apm_ruby.git', branch: 'scoutprof_statsd_context'
```

```
diff --git a/Gemfile.lock b/Gemfile.lock
index 353ce16..a4f3768 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,9 +1,9 @@
 GIT
   remote: https://github.com/scoutapp/scout_apm_ruby.git
-  revision: 634484b7673baf1083b8b3076118c100ab2e022e
-  branch: scoutprof_statsd_context
+  revision: 3f469a854bf57d23fd778e9860628eceabd64196
+  branch: auto_instruments
   specs:
-    scout_apm (3.0.0.pre23)
+    scout_apm (2.4.11)
 
 GEM
   remote: https://rubygems.org/
@@ -51,6 +51,7 @@ GEM
     airbrussh (1.1.0)
       sshkit (>= 1.6.1, != 1.7.0)
     arel (6.0.3)
+    ast (2.4.0)
     autoprefixer-rails (6.5.0.1)
       execjs
     aws-sdk (2.0.28)
@@ -268,6 +269,9 @@ GEM
     oj (3.1.3)
     orm_adapter (0.5.0)
     os (0.9.6)
+    parallel (1.12.1)
+    parser (2.5.1.0)
+      ast (~> 2.4.0)
     pg (0.18.3)
     pg-eyeballs (1.1.0)
       activerecord (>= 4.0, < 6.0)
@@ -280,6 +284,7 @@ GEM
       capybara (~> 2.1)
       cliver (~> 0.3.1)
       websocket-driver (>= 0.2.0)
+    powerpack (0.1.1)
     premailer (1.8.7)
       css_parser (>= 1.4.5)
       htmlentities (>= 4.0.0)
@@ -334,6 +339,7 @@ GEM
       activesupport (= 4.2.7.1)
       rake (>= 0.8.7)
       thor (>= 0.18.1, < 2.0)
+    rainbow (3.0.0)
     raindrops (0.17.0)
     rake (11.3.0)
     rbnacl (3.4.0)
@@ -373,6 +379,14 @@ GEM
     rollbar (2.15.5)
       multi_json
     rouge (1.10.1)
+    rubocop (0.56.0)
+      parallel (~> 1.10)
+      parser (>= 2.5)
+      powerpack (~> 0.1)
+      rainbow (>= 2.2.2, < 4.0)
+      ruby-progressbar (~> 1.7)
+      unicode-display_width (~> 1.0, >= 1.0.1)
+    ruby-progressbar (1.9.0)
     rugged (0.23.2)
     safe_yaml (1.0.4)
     sass (3.4.22)
@@ -476,6 +490,7 @@ GEM
       unf_ext
     unf_ext (0.0.7.1)
     unicode (0.4.4.2)
+    unicode-display_width (1.3.3)
     unicorn (5.2.0)
       kgio (~> 2.6)
       raindrops (~> 0.7)
@@ -558,6 +573,7 @@ DEPENDENCIES
   retriable (~> 2.1)
   rollbar
   rouge
+  rubocop
   rugged
   sass-rails
   scout_apm!
```

```
diff --git a/config/boot.rb b/config/boot.rb
index 6b750f0..961b641 100644
--- a/config/boot.rb
+++ b/config/boot.rb
@@ -1,3 +1,3 @@
 ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile', __FILE__)
-
+require 'scout_apm/auto_instrument'
 require 'bundler/setup' # Set up gems listed in the Gemfile.
```

itsderek23:
Also:

```
ruby -v
ruby 2.1.2p95 (2014-05-08 revision 45877) [x86_64-darwin13.0]
```

I noticed that `ScoutApm::InstructionSequence#load_iseq` isn't called. I wonder if this is due to my older Ruby version.

itsderek23:
Works with Ruby 2.3.1, so a Ruby version issue. 

Pretty cool - this breaks down a lot of the time!

image

itsderek23:
This helps address https://github.com/scoutapp/roadmap/issues/68

dlanderson:
Correct - needs Ruby 2.3.1+ currently.

ioquatix:
Okay, I've made a first stab at improving this.

I've added some tests.

I suggest that the next step is to ensure we are generating the correct instrumentation and add a handful more tests.

ioquatix:
The code right now is very much Rails specific, but I've left it open for the future, we can probably modify it to suit different frameworks e.g. Sinatra.

ioquatix:
Do you have any prescribed method for doing integration tests? i.e. create a default rails site, insert the `scout_apm` gem, run it, and check that it's instrumented/collecting data?

I feel like this would be a nice addition.

Gemfile
```diff
@@ -10,3 +10,7 @@ if RUBY_VERSION <= "1.8.7"
   gem "pry", "~> 0.9.12"
   gem "rake", "~> 10.5"
 end
+
+group :development do
+  gem "parser"
```
cschneid: Can this be a dev-time gemspec dependency instead?

lib/scout_apm/layer_children_set.rb
```diff
@@ -69,6 +69,14 @@ def each
       end
     end
 
+    def length
+      @children.length
+    end
+
+    def size
+      @children.size
+    end
```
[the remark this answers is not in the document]
ioquatix: @dlanderson made this change so I can't really comment on why it was done. I can investigate in more detail.

lib/scout_apm/auto_instrument.rb
```diff
@@ -0,0 +1,2 @@
+
+require 'scout_apm/auto_instrument/instruction_sequence'
```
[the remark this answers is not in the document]
ioquatix: I assume for customers we need to tell them something simple like "Add `require 'scout_apm/auto_instrument`". We might change the layout of files under the directory, but this should be the top level entry point for this functionality. We can certainly tell them to include the instruction sequence file directly, but that might be implementation-specific, e.g. in the future we might wrap `require` or use `TracePoint`.

Gemfile
ioquatix: Yes, no problem.

lib/scout_apm/layer_children_set.rb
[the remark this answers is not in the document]
dlanderson: because `Layer#to_s` (used for debugging/printing in `auto_isntruments` calls it's own `@children.length` and it's not defined.

lib/scout_apm/auto_instrument/instruction_sequence.rb
```diff
@@ -0,0 +1,22 @@
...
+module ScoutApm
+  module AutoInstrument
+    module InstructionSequence
+      CONTROLLER_PATH_PATTERN = /\/apm\/app\/controllers\/.*_controller.rb$/
+
+      def load_iseq(path)
+        if path =~ CONTROLLER_PATH_PATTERN
+          new_code = Rails.rewrite(path)
+
+          return compile(new_code, path, filepath)
```
dlanderson: `filepath` is undefined. Also, I think the arguments to `compile` are wrong - [https://ruby-doc.org/core-2.5.0/RubyVM/InstructionSequence.html#method-c-compile](https://ruby-doc.org/core-2.5.0/RubyVM/InstructionSequence.html#method-c-compile)

lib/scout_apm/auto_instrument/instruction_sequence.rb
```diff
@@ -0,0 +1,22 @@
+
+require 'scout_apm/auto_instrument/rails'
+
+module ScoutApm
+  module AutoInstrument
+    module InstructionSequence
+      CONTROLLER_PATH_PATTERN = /\/apm\/app\/controllers\/.*_controller.rb$/
```
dlanderson: We need to make sure we remove the `\/apm` portion here

lib/scout_apm/auto_instrument/instruction_sequence.rb
```diff
@@ -0,0 +1,22 @@
...
+          new_code = Rails.rewrite(path)
+
+          return compile(new_code, path, filepath)
+        else
+          return super
+        end
+      end
+    end
+
+    RubyVM::InstructionSequence.prepend(InstructionSequence)
```
dlanderson: Oddly, this is prepending `ScoutAPM::AutoInstrument::InstructionSequence` to `RubyVM::InstructionSequence`, but `load_iseq` does not get defined. Let's chat together on why that might be

lib/scout_apm/auto_instrument/rails.rb
```diff
@@ -0,0 +1,73 @@
+
+# In order for this to work, you must add `gem 'parser'` to your Gemfile.
+require 'parser/current'
```
dlanderson: Thinking we'll be more defensive here and use conditionals around our definitions and code in order to prevent things blowing up if someone doesn't include `parser` or we're running in an older ruby version, non MRI, etc.

lib/scout_apm/auto_instrument/rails.rb
```diff
@@ -0,0 +1,73 @@
...
+        buffer = Parser::Source::Buffer.new(path)
+        buffer.source = code
+
+        rewriter = Rewriter.new
+
+        # Rewrite the AST, returns a String with the new form.
+        rewriter.rewrite(buffer, ast)
+      end
+
+      class Rewriter < ::Parser::TreeRewriter
```
dlanderson: Going to need you to walk me through the logic used in this class just so I understand fully :)

lib/scout_apm/auto_instrument/rails.rb
```diff
@@ -0,0 +1,73 @@
...
+        # end
+
+        def on_send(node)
+          return if parent_type?(:block) or @method.empty?
+
+          line = node.location.line || 'line?'
+          column = node.location.column || 'column?'
+          method_name = node.children[1] || '*unknown*'
+
+          wrap(node.location.expression, "::ScoutApm::Instruments::AutoInstruments.dynamic_layer('#{method_name}:l#{line}:c#{column}'){", "}")
```
dlanderson: `::ScoutApm::Instruments::AutoInstruments.dynamic_layer` si so... ugly 😜 Thinking about how we can make that better. We mentioned about the invisible space or hot dog. Need to decide on options here

lib/scout_apm/instruments/auto_instruments.rb
```diff
@@ -0,0 +1,25 @@
...
+  module Instruments
+    class AutoInstruments
+      def self.dynamic_layer(name)
+        req = ScoutApm::RequestManager.lookup
+
+        begin
+          layer = ScoutApm::Layer.new('AutoInstrument', name)
+          req.start_layer(layer)
+          started_layer = true
+          puts "================ STARTED AUTO INSTRUMENT LAYER #{name} ============="
```
dlanderson: Reminder to remove this debug code before releasing to customers

test/unit/auto_instrument_test.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def instrumented_source(name)
+    File.read(instrumented_path(name))
+  end
+
+  def test_rails_controller_rewrite
+    assert_equal instrumented_source("controller"), ::ScoutApm::AutoInstrument::Rails.rewrite(source_path("controller"))
+  end
+end if RUBY_VERSION >= "2.3.1"
```
dlanderson: I think we can test the rewriter on all versions. It's the `load_iseq` feature that's version dependent

lib/scout_apm/auto_instrument/instruction_sequence.rb
```diff
@@ -0,0 +1,22 @@
...
+    module InstructionSequence
+      CONTROLLER_PATH_PATTERN = /\/apm\/app\/controllers\/.*_controller.rb$/
+
+      def load_iseq(path)
+        if path =~ CONTROLLER_PATH_PATTERN
+          new_code = Rails.rewrite(path)
+
+          return compile(new_code, path, filepath)
+        else
+          return super
```
dlanderson: Forgot this crucial one: we must assume `super` (`load_iseq`) does not exist (but it *might* exist).

lib/scout_apm/auto_instrument/rails.rb
ioquatix: Okay I wrapped it and `rescue LoadError`. Let's see how this works on travis.

lib/scout_apm/auto_instrument/rails.rb
ioquatix: I implemented this change in recent commit.

lib/scout_apm/auto_instrument/instruction_sequence.rb
ioquatix: Yeah, I reworked it to be `compile_file`. It's simple solution for now.

ioquatix: We could probably write `super rescue compile_file(path)` but that might be a bad idea.

lib/scout_apm/auto_instrument/instruction_sequence.rb
```diff
@@ -0,0 +1,30 @@
...
+        
+        return self.compile_file(path)
+      end
+    end
+    
+    # This should work (https://bugs.ruby-lang.org/issues/15572), but it doesn't.
+    # RubyVM::InstructionSequence.extend(InstructionSequence)
+
+    # So we do this instead:
+    class << ::RubyVM::InstructionSequence
```
dlanderson: @ioquatix Can you put conditionals around this to prevent trying to prepend on older ruby versions? I think 2.3.1+ is the minimum.

lib/scout_apm/auto_instrument.rb
```diff
@@ -0,0 +1,16 @@
...
+		raise LoadError, "Parser::TreeRewriter was not defined" unless defined?(Parser::TreeRewriter)
+		
+		require 'scout_apm/auto_instrument/instruction_sequence'
+		require 'scout_apm/auto_instrument/layer'
+	rescue LoadError
+		warn "ScoutApm::AutoInstrument requires `gem 'parser'` to be present: #{$!}. Skipping."
+	end
+else
+	warn "ScoutApm::AutoInstrument requires Ruby >= v2.3. Skipping."
+end
```
cschneid: 👮 hard tabs?

lib/scout_apm/auto_instrument/rails.rb
```diff
@@ -0,0 +1,130 @@
...
+
+          # Keeps track of the parent - child relationship between nodes:
+          @nesting = []
+
+          # The stack of method nodes (type :def):
+          @method = []
+        end
+
+        def instrument(method_name, line, column)
+          ["::ScoutApm::AutoInstrument(\"\#{self.class}\\\##{method_name}:L#{line}:C#{column}\"){", "}"]
```
cschneid: The `desc` field of a layer may make sense to put an exact line & column number in. 

So the name can be `Foo/length` and desc is `line 3, column 8` or whatever format you want.

lib/scout_apm/auto_instrument/layer.rb
```diff
@@ -0,0 +1,21 @@
...
+    request = ScoutApm::RequestManager.lookup
+
+    begin
+      layer = ScoutApm::Layer.new('AutoInstrument', name)
+      request.start_layer(layer)
+      started_layer = true
+
+      result = yield
+    rescue
+      request.error!
```
cschneid: `error!` should only be marked true if the controller lets an exception escape, but this will mark a request as an error if any individual call exceptions, even if caught right away.

If the controller layer still exists, we should let it handle this error! mark.

During the quick test on APM prod, some code like below was being marked as an "error" due to this.

```
def index
  something_that_raises
rescue
  render text: "ok"
end
```

lib/scout_apm/auto_instrument.rb
ioquatix: Okay I fixed it. My bad.

lib/scout_apm/auto_instrument/layer.rb
ioquatix: That makes sense. So perhaps we shouldn't handle errors and just propagate them up?

lib/scout_apm/auto_instrument/rails.rb
ioquatix: The latest version of this puts the class and controller method with line number into the name field and the first line of code/expression into the description field. How do you feel about that? Can you try it and see if it makes sense?

lib/scout_apm/auto_instrument/layer.rb
ioquatix: Okay done.

jsierles:
Thanks for the work on this!

Just one finding. On Heroku, the Rails console is not run using bundler, so requiring in `boot.rb` would fail. To remedy this, I added the gem to a `pre_boot` bundler group:

```
require "bundler"
Bundler.setup(:pre_boot)

require 'scout_apm/auto_instrument'
```

itsderek23:
Should we set the version to something like `2.4.21.pre.auto`? Without that, I'm unsure how to check which apps are running this branch.

dlanderson:
Replaced by https://github.com/scoutapp/scout_apm_ruby/pull/247

PR #200, opened 2018-06-21
Incorporating total time consumed (response time * throughput) into trace policy

itsderek23:
Our current trace policy looks at a number of factors to determine if we should store a transaction trace. One it did not look at is the total time consumed of endpoints, which is the default sort order of our endpoints.

This incorporates total time consumed into our formula.


itsderek23:
See https://github.com/scoutapp/roadmap/issues/70#issuecomment-398556207

itsderek23:
I'm seeing better results w/this testing against a production-like traffic flow.

The data set I'm using comes from a real application. "index0" is the most time consuming endpoint. "index231" is the least.

There's a large dropoff after the first 20 or so endpoints.

With the updated trace policy, 5 of the 10 endpoints w/the most time consumed are among the top 10 endpoints in terms of number of traces collected:

image

It's not clear to me why endpoints like `index77` and `index61` made the top 10. That said, it's still a huge step in the right direction. With the current trace policy, zero made the list:

image

@cschneid - I'm in favor of doing a new release w/this update. It appears to be a nice upgrade.

PR #204, opened 2018-07-17
ActiveRecord query naming enhancements

itsderek23:
`BEGIN` and `COMMIT` can take significant time. This identifies these queries in the metric name.

Currently, these are named `SQL/other`, which is our fallback when we're unable to come up with a label for an SQL query.

### DIFF
--- a/lib/scout_apm/layer_converters/database_converter.rb
+++ b/lib/scout_apm/layer_converters/database_converter.rb
@@ -13,7 +13,6 @@ module ScoutApm
 
         walker.on do |layer|
           next if skip_layer?(layer)
-
           stat = DbQueryMetricStats.new(
             model_name(layer),
             operation_name(layer),
@@ -52,23 +51,11 @@ module ScoutApm
       DEFAULT_OPERATION = "other"
 
       def model_name(layer)
-        if layer.name.respond_to?(:model)
-          layer.name.model || DEFAULT_MODEL
-        else
-          DEFAULT_MODEL
-        end
-      rescue
-        DEFAULT_MODEL
+        layer.name.to_s.split("/").first || DEFAULT_MODEL
       end
 
       def operation_name(layer)
-        if layer.name.respond_to?(:normalized_operation)
-          layer.name.normalized_operation || DEFAULT_OPERATION
-        else
-          DEFAULT_OPERATION
-        end
-      rescue
-        DEFAULT_OPERATION
+        layer.name.to_s.split("/")[1] || DEFAULT_OPERATION
       end
 
       def records_returned(layer)
--- a/lib/scout_apm/utils/active_record_metric_name.rb
+++ b/lib/scout_apm/utils/active_record_metric_name.rb
@@ -130,7 +130,7 @@ module ScoutApm
             else
               SELECT_LABEL
             end
-          "#{match[3].gsub!(/\W/,'').classify}/#{operation}"
+          "#{match[3].gsub(/\W/,'').classify}/#{operation}"
         elsif match = UPDATE_REGEX.match(sql)
           "#{match[2].classify}/#{UPDATE_LABEL}"
         elsif match = INSERT_REGEX.match(sql)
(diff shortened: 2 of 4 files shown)


lib/scout_apm/layer_converters/database_converter.rb
```diff
@@ -52,23 +51,11 @@ def skip_layer?(layer)
...
 
       def model_name(layer)
-        if layer.name.respond_to?(:model)
-          layer.name.model || DEFAULT_MODEL
-        else
-          DEFAULT_MODEL
-        end
-      rescue
-        DEFAULT_MODEL
+        layer.name.to_s.split("/").first || DEFAULT_MODEL
```
itsderek23: `model` is nil for layers that don't have a name and just SQL. This instead takes the output of `to_s` to generate the final model name.

lib/scout_apm/layer_converters/database_converter.rb
```diff
@@ -52,23 +51,11 @@ def skip_layer?(layer)
...
 
       def operation_name(layer)
-        if layer.name.respond_to?(:normalized_operation)
-          layer.name.normalized_operation || DEFAULT_OPERATION
-        else
-          DEFAULT_OPERATION
-        end
-      rescue
-        DEFAULT_OPERATION
+        layer.name.to_s.split("/")[1] || DEFAULT_OPERATION
```
itsderek23: `operation` is nil for layers that don't have a name and just SQL. This instead takes the output of `to_s` to generate the final operation name.

itsderek23:
This PR also addresses another problem where our named DB queries don't match what we display in transaction traces. 

Previously, `DatabaseConverter` would generate a metric name based on the `ActiveRecordName`'s `model` and `operation`. However, these are both `nil` if we aren't provided with a label from ActiveRecord and need to manually generate a name by parsing the SQL. This means that queries w/o specific labels are identified as just `SQL/other` in the database area of Scout. Additionally, since the metrics are named differently, they would frequently display zero actual queries occurring in traces.

lib/scout_apm/utils/active_record_metric_name.rb
```diff
@@ -121,13 +130,17 @@ def regex_name(sql)
             else
               SELECT_LABEL
             end
-          "#{match[3].classify}/#{operation}"
+          "#{match[3].gsub(/\W/,'').classify}/#{operation}"
```
cschneid: What's the purpose of the new gsub? It strips non-word-chars right?

lib/scout_apm/utils/active_record_metric_name.rb
```diff
@@ -86,6 +93,8 @@ def parse_operation
       NON_GREEDY_CONSUME = '.*?'
       TABLE = '(?:"|`)?(.*?)(?:"|`)?\s'
       COUNT = 'COUNT\(.*?\)'
+      BEGIN_STATEMENT = 'BEGIN'.freeze # BEGIN is a reserved keyword
+      COMMIT = 'COMMIT'.freeze
```
cschneid: We should standardize the name of the constant between the two (`_STATEMENT` or not).

Also these constants make more sense down below with the `_LABEL` constants I think. 

Or we could just fold it directly into the code where it's used.

lib/scout_apm/layer_converters/database_converter.rb
cschneid: This is better than the cascading if/else/rescue that was there before, but still seems roundabout to do a full `.to_s` only to then split it back out.  I'll think about how that could be better.

lib/scout_apm/utils/active_record_metric_name.rb
itsderek23: Yes - I saw some queries names with a `)`, for example.

cschneid: Ah, hmm - should we extend the underlying regex for it vs. adding the special case here?

itsderek23:
Released in:

* 2.4.17
* 3.0.0.pre27

Issue #211, opened 2018-08-03
Failed writing data to layaway file: no _dump_data is defined for class Thread::Mutex`


bjeanes:
In upgrading a Rails application from Rails 4.1 to 4.2, Scout stops sending data entirely. The logs contain (I've formatted the backtrace-as-array that is printed):

```
Failed writing data to layaway file: no _dump_data is defined for class Thread::Mutex / 
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/layaway_file.rb:32:in `dump'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/layaway_file.rb:32:in `serialize'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/layaway_file.rb:27:in `write'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/layaway.rb:61:in `write_reporting_period'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/store.rb:102:in `block in write_reporting_period'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/store.rb:101:in `synchronize'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/store.rb:101:in `write_reporting_period'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/store.rb:91:in `block in write_to_layaway'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/store.rb:91:in `each'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/store.rb:91:in `write_to_layaway'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/reporting.rb:31:in `process_metrics'
/app/vendor/bundle/ruby/2.3.0/gems/scout_apm-2.4.14/lib/scout_apm/periodic_work.rb:13:in `run'
[...]

PR #212, opened 2018-08-04
Only annotate layer with known payload

bjeanes:
Potential fix for #211 

Datadog mutates the `:payload` key in the ActiveSupport instrumentation used by both Scout and Datadog, resulting in Scout encountering unexpected data, which it then annotates to layer and eventually attempts to dump to the layaway file. Dumping fails due to the annotations containing un-marshalable `Thread::Mutex`es.

This proposes being stricter in what keys to copy from ActiveRecord instrumentation payloads and solves the problem locally for me. In addition to opening up compatibility with Datadog, it's arguably more future-proof to other types of data being added here that we may not care about.

I'm on Rails 4.2, so it may be that later Rails versions have already added keys here that Scout APM finds useable, in which case this would need to be revised.

If this approach is acceptable, please let me know if you'd like a separate PR for the v3 line.

### DIFF
--- a/lib/scout_apm/instruments/active_record.rb
+++ b/lib/scout_apm/instruments/active_record.rb
@@ -142,7 +142,10 @@ module ScoutApm
             req = ScoutApm::RequestManager.lookup
             layer = req.current_layer
             if layer && layer.type == "ActiveRecord"
-              layer.annotate_layer(payload)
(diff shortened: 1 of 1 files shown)


dlanderson:
Thanks, @bjeanes - this is exactly the issue. Thanks for investigating!

PR #213, opened 2018-08-04
Only annotate layer with known payload (scoutprof)

bjeanes:
Potential fix for #211 for scoutprof. See #212 for v2 fix


dlanderson:
Closed via #212 (which has been merged into scoutprof branch from master). Thanks again @bjeanes !

PR #215, opened 2018-08-22
Initial Shoryuken support

### LINKED ISSUE #195: Shoryuken instrumentation
I would appreciate instrumentation for the Amazon SQS background job worker, [Shoryuken](https://github.com/phstc/shoryuken).

It has [middleware](https://github.com/phstc/shoryuken/wiki/Middleware) based on Sidekiq.

cschneid:
If Shoryuken is launched as part of Rails, it will be detected automatically.

If Shoryuken is launched separately, as a standalone worker, an initializer is
required to start the agent:

```ruby
Shoryuken.configure_server do |config|
  config.on(:startup) do
    require 'scout_apm'
    ScoutApm::Agent.install
    ScoutApm::Agent.start
  end
end
```

Closes #195

### DIFF
--- a/lib/scout_apm.rb
+++ b/lib/scout_apm.rb
@@ -58,6 +58,7 @@ require 'scout_apm/server_integrations/null'
 require 'scout_apm/background_job_integrations/sidekiq'
 require 'scout_apm/background_job_integrations/delayed_job'
 require 'scout_apm/background_job_integrations/resque'
+require 'scout_apm/background_job_integrations/shoryuken'
 
 require 'scout_apm/framework_integrations/rails_2'
 require 'scout_apm/framework_integrations/rails_3_or_4'
--- a/lib/scout_apm/background_job_integrations/shoryuken.rb
+++ b/lib/scout_apm/background_job_integrations/shoryuken.rb
@@ -56,7 +56,16 @@ module ScoutApm
     # and time them.
     class ShoryukenMiddleware
       def call(worker_instance, queue, msg, body)
-        job_class = worker_instance.class.to_s
+        job_class =
+          begin
+            if worker_instance.class.to_s == ACTIVE_JOB_KLASS
+              body["job_class"]
+            else
+              worker_instance.class.to_s
+            end
+          rescue
+            UNKNOWN_CLASS_PLACEHOLDER
+          end
 
         req = ScoutApm::RequestManager.lookup
         req.annotate_request(:queue_latency => latency(msg))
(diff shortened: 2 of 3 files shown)


tzar:
I have tried this, but all jobs seem to come through as `ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper` rather than the actual name of the job class. Looking at the sidekiq integration (as it is quite similar), it looks like the resolution of `job_class` needs to be a bit more complicated:

https://github.com/scoutapp/scout_apm_ruby/blob/6bf98166f885ebd57ee9051c719519f636d4525d/lib/scout_apm/background_job_integrations/sidekiq.rb#L83

cschneid:
@tzar thank you for trying that - I swear that worked in my ActiveJob test setup - I'll go double check and tweak that naming.

cschneid:
@tzar - a quick check shows that ActiveJob based execution seems to work the way I have originally. I wonder if I'm launching Shoryuken differently than you are, or perhaps a different version? Can you see what doesn't match up between our environments and I'll reproduce what you're seeing.

Version: `shoryuken 3.2.3`

Launching with `bundle exec shoryuken -q shoryuken_test -R`

The job itself is trivial:

```ruby
class DummyWorker
  include Shoryuken::Worker

  shoryuken_options queue: 'shoryuken_test', auto_delete: true

  def perform(sqs_msg, name)
    puts "Hello, #{name}"
  end
end
```

tzar:
@cschneid I suspect the difference is that our shoryuken jobs look like this:

```ruby
class ApplicationJob < ActiveJob::Base
  queue_as :default
end
class DummyJob < ApplicationJob
  def perform(*args)
    pp args
  end
end
```
We're also running shoryuken 3.2.3 and launching with `bundle exec shoryuken -C config/shoryuken-worker.yml -R`

config file is pretty simple:

```yaml
delay: <%= ENV["SQS_DELAY"] %>
concurrency: 6
queues:
  - [default, 8]
```
To connect active job to shoryuken, you'll also need:
```
config/application.rb:    config.active_job.queue_adapter = :shoryuken
```
Hope this helps!

cschneid:
@tzar Thank you for the clarification, I bet the inheritance from ActiveJob is the difference. I'll get this PR adapted to that.

cschneid:
@tzar ok - I reproduced your setup locally and I added a commit to capture the name correctly now. Can you give it a shot?

tzar:
@cschneid the job classes are coming through correctly! Thanks for your work on this, it is much appreciated :smile:

PR #229, opened 2018-11-15
Ability to disable sending remote ip

AndriiAkulov:


AndriiAkulov:
https://github.com/scoutapp/scout_apm_ruby/pull/229

README.markdown
```diff
@@ -55,6 +55,13 @@ gem 'scout_apm'
 SCOUT_DEV_TRACE=true rails server
 ```
 
+## How to test gem locally
```
cschneid: Newline after this line.

README.markdown
```diff
@@ -55,6 +55,13 @@ gem 'scout_apm'
...
+If you want to make some changes in the gem code and test it locally, then you need to specify in Gemfile of your application
+the path to your gem. `gem 'scout_apm', path: 'path_to_your_gem/scout_apm_ruby'`. Also, you need manually
+compile native extensions: 
+'bundle exec rake compile' in the root directory of `scout_apm_ruby` gem.
```
cschneid: Probably better as a bulleted list:

```
* Point your gemfile at the local checkout: `gem 'scout_apm', path: '/path/to/scout_apm_ruby`
* Compile native code: `cd /path/to/scout_apm_ruby && bundle exec rake compile`
```

lib/scout_apm/config.rb
```diff
@@ -268,6 +268,7 @@ class ConfigDefaults
         'database_metric_report_limit' => 1000,
         'instrument_http_url_length' => 300,
         'start_resque_server_instrument' => true, # still only starts if Resque is detected
+        'collect_remote_ip' => true
```
cschneid: Leave a trailing comma to minimize future diffs (when we add another option).

lib/scout_apm/instruments/action_controller_rails_2.rb
```diff
@@ -6,6 +6,7 @@ class ActionControllerRails2
       def initialize(context)
         @context = context
         @installed = false
+        @agent_context = ScoutApm::Agent.instance.context
```
cschneid: context is passed in right above, in the `@context` variable. And it doesn't look like this instance variable is even used?

lib/scout_apm/instruments/action_controller_rails_2.rb
```diff
@@ -38,7 +39,8 @@ def install
     module ActionControllerRails2Instruments
       def self.included(instrumented_class)
         # XXX: Don't lookup context by global
-        ScoutApm::Agent.instance.context.logger.info "Instrumenting #{instrumented_class.inspect}"
+        @agent_context = ScoutApm::Agent.instance.context
+        @agent_context.logger.info "Instrumenting #{instrumented_class.inspect}"
```
cschneid: Don't create an instance variable here - we're inside the context of customer's controllers at this point, and we need to be careful not to pollute their namespace with extra instance variables.

Looking up by the global is fine and doesn't need to be changed.

AndriiAkulov:
@cschneid Fixed your comments and repushed.

lib/scout_apm/config.rb
```diff
@@ -268,6 +268,7 @@ class ConfigDefaults
         'database_metric_report_limit' => 1000,
         'instrument_http_url_length' => 300,
         'start_resque_server_instrument' => true, # still only starts if Resque is detected
+        'collect_remote_ip' => true,
```
cschneid: Add this to the boolean coercions like #234 did

PR #236, opened 2018-12-03
Fix embedded Help Doc URLs

cschneid:


cschneid:
No idea why this sat for so long, but merging....

PR #259, opened 2019-03-24
Add monitor false for test env

abhishek77in:
Without setting the test env, we get an error:
```
Couldn't find configuration in app/config/scout_apm.yml for environment: test.
```


cschneid:
Thank you!

Issue #258, opened 2019-03-21
Scout APM does not appear to consider http.rb time as HTTP


ibrahima:
Hi! I use [HTTP.rb](https://github.com/httprb/http) as an HTTP client in some places because it's nicer than other clients in some ways, and it seems that Scout does not track time spent in that library as HTTP. Instead it appears to consider that time as spent in the view or the controller depending on where the method was that triggered an HTTP call. It'd be great to get this accounting fixed. I am using scout_apm (3.0.0.pre28).

Thanks!

ioquatix:
I will instrument `HTTP.rb` similar to https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/instruments/net_http.rb

ibrahima:
That would be awesome, thank you!

jonekdahl:
We're switching to http.rb and it would be great to have Scout instrumentation in place. :+1:

jrothrock:
Reopening as we aren't instrumenting HTTP.rb calls in the latest versions of Scout and HTTP.rb

jrothrock:
Closing this issue in favor of the above

PR #260, opened 2019-03-31
Add HTTP instrumentation.

ioquatix:
Fixes https://github.com/scoutapp/scout_apm_ruby/issues/258

This requires testing. I'm not aware of how this should be done, but how about something like this?

```ruby
def test_http_instrumentation
	context = ScoutApm::TestContext.new
	
	ScoutApm::Instruments::HTTP.install(context)
	
	events = context.track_events do
		HTTP.get("http://www.google.com/search?q=kittens")
	end
	
	assert_equal events, [
		["HTTP", "get", "http://www.google.com/search"]
	]
end
```

cc @dlanderson @cschneid

### DIFF
--- a/lib/scout_apm/instruments/http.rb
+++ b/lib/scout_apm/instruments/http.rb
@@ -34,6 +34,8 @@ module ScoutApm
             def request_scout_description(verb, uri)
               max_length = ScoutApm::Agent.instance.context.config.value('instrument_http_url_length')
               (String(uri).split('?').first)[0..(max_length - 1)]
+            rescue
+              ""
             end
 
             alias request_without_scout_instruments request
--- a/lib/scout_apm/instruments/http.rb
+++ b/lib/scout_apm/instruments/http.rb
@@ -20,7 +20,7 @@ module ScoutApm
         if defined?(::HTTP) && defined?(::HTTP::Client)
           @installed = true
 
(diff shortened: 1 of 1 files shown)


jonekdahl:
Is this still blocked on lack of tests? If not, we'd really like to see this merged.

ioquatix:
Okay, I will review it this weekend and merge it.

jonekdahl:
Hi @ioquatix, I don't want you to rush this, but please let me know if you need help testing this or something. Thanks!

lib/scout_apm/instruments/http.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      def installed?
+        @installed
+      end
+
+      def install
+        if defined?(::HTTP) && defined?(::HTTP::Client)
+          @installed = true
+
+          logger.info "Instrumenting HTTP"
```
dlanderson: Change to "Instrumenting HTTP::Client" since we already have 2 other HTTP instruments ([http_client](https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/instruments/http_client.rb) and [net_http](https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/instruments/net_http.rb))

PR #262, opened 2019-04-22
Postgres Sanitization doesn't handle literal strings in SELECT clause

bohdanbenov:
Closes scoutapp/apm#553

### DIFF
--- a/lib/scout_apm/utils/sql_sanitizer.rb
+++ b/lib/scout_apm/utils/sql_sanitizer.rb
@@ -50,7 +50,7 @@ module ScoutApm
       def to_s_postgres
         sql.gsub!(PSQL_PLACEHOLDER, '?')
         sql.gsub!(PSQL_VAR_INTERPOLATION, '')
-        sql.gsub!(PSQL_REMOVE_STRINGS, '?')
+        sql.gsub!(PSQL_AFTER_WHERE) {|c| c.gsub(PSQL_REMOVE_STRINGS, '?')}
         sql.gsub!(PSQL_REMOVE_INTEGERS, '?')
         sql.gsub!(PSQL_IN_CLAUSE, 'IN (?)')
         sql.gsub!(MULTIPLE_SPACES, ' ')
--- a/lib/scout_apm/utils/sql_sanitizer_regex.rb
+++ b/lib/scout_apm/utils/sql_sanitizer_regex.rb
@@ -11,6 +11,7 @@ module ScoutApm
       PSQL_REMOVE_INTEGERS = /(?<!LIMIT )\b\d+\b/.freeze
       PSQL_PLACEHOLDER = /\$\d+/.freeze
       PSQL_IN_CLAUSE = /IN\s+\(\?[^\)]*\)/.freeze
+      PSQL_AFTER_WHERE = /(?<=where ).*$/i.freeze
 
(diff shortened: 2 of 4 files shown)


lib/scout_apm/utils/sql_sanitizer.rb
```diff
@@ -42,7 +42,7 @@ def to_s
       def to_s_postgres
         sql.gsub!(PSQL_PLACEHOLDER, '?')
         sql.gsub!(PSQL_VAR_INTERPOLATION, '')
-        sql.gsub!(PSQL_REMOVE_STRINGS, '?')
+        sql.gsub!(PSQL_AFTER_WHERE) {|c| c.gsub(PSQL_REMOVE_STRINGS, '?')}
```
cschneid: I don't like this name. Every other name is the "what" of the regex, and this name is the "where". Can you reuse the original `PSQL_REMOVE_STRINGS` name instead?

cschneid: `gsub` is probably wrong. I don't think that an SQL query can have multiple where clauses?

lib/scout_apm/utils/sql_sanitizer_regex.rb
```diff
@@ -10,6 +10,7 @@ module SqlRegex
       PSQL_REMOVE_INTEGERS = /(?<!LIMIT )\b\d+\b/.freeze
       PSQL_PLACEHOLDER = /\$\d+/.freeze
       PSQL_IN_CLAUSE = /IN\s+\(\?[^\)]*\)/.freeze
+      PSQL_AFTER_WHERE = /(?<=where ).*$/i.freeze
```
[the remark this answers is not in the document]
cschneid: I don't think the lookbehind is needed in the regex, just adds complexity. Since it just gets sent through the string sanitizer anyway, which will never match the 'WHERE' part.

lib/scout_apm/utils/sql_sanitizer.rb
dlanderson: Subqueries?

cschneid: Valid - but the PR code doesn't work w/ subqueries either.

bohdanbenov: > I don't like this name. Every other name is the "what" of the regex, and this name is the "where". Can you reuse the original `PSQL_REMOVE_STRINGS` name instead?

@cschneid By reusing the original `PSQL_REMOVE_STRINGS` name do you mean to rename this regex to something like `PSQL_REMOVE_STRINGS_AFTER_WHERE`? Or unite "after where" regex with  "remove strings" regex, so   `PSQL_REMOVE_STRINGS` will do both: checks "after where" and then check literal strings?

cschneid: The name is fine after reading the code a bit more, but do add tests to make sure this doesn't break with subqueries, and also that this code fixes the original issue.

bohdanbenov: I added the tests for strings and change regex for the subqueries. Regex do check from `where` statement to the `select` of a subquery or to the end on the string
4711abcc_2019 25 04
Here is a screenshot of the query from the issue 
e03d2747_2019 25 04

PR #265, opened 2019-06-06
Background job integrations/que

ioquatix:


ioquatix:
@nearapogee can you explain whether we need to expand the current implementation to support the latest release of `que` or are you happy with this as it is?

nearapogee:
0.14.3 is the latest release. 1.0.0 I think has major changes, but isn't officially out yet.

I would be happy to create another pull request once 1.0.0 comes out with the necessary changes. We will be adopting it rather quickly, I believe. I just don't want to do that until I know the 1.0.0 api is stable, probably is, but work is busy....

If I were you I would put this out, knowing/stating it only supports currently released versions of que. Does that fit with your philosophy? What do you recommend?

ioquatix:
I personal think that approach is fine. I'll review it in more detail tomorrow and hopefully we can merge.

nearapogee:
Sounds good. If I remember right, I saw other implementations with version checks as well.

ioquatix:
@nearapogee I've put a check around `::Que::Version` so that we only load the Scout instrumentation for que `0.x` releases for now. Once all specs pass, I'll merge into master. Do you think you can try it and let me know if it's working for you?

ioquatix:
It looks like que might be dead.

https://github.com/chanks/que/issues/245

jordoh:
> It looks like que might be dead.

I've had 1.0.0 beta deployed in multiple production applications for quite a while now, with no issues. `scout_apm` support would definitely still be appreciated, even if 1.0.0 stays in "beta" perpetually.

nearapogee:
@ioquatix I'll update and check.

I agree with @jordoh on having the scout implementation if possible. I like que enough that I may be interested in maintaining it. Just saw this and will have to look into it though.

PR #269, opened 2019-07-03
Emulate prepend on 1.x, direct support on 2.x

ioquatix:
On 1.x it is possible to simulate prepend. For a given model containing methods to be prepended, firstly we `include` an anonymous module. Then, for all the methods named, we move the implementation from the class into the anonymous module. Then we add the methods from the extension module to the class. The extension module, on calling super, will invoke the original implementation in the module which is now above it in the `ancestors` chain. This is better than `alias_method` because no new methods are introduced, and it's also for all practical purposes, identical to using prepend.

### DIFF
--- a/lib/scout_apm/extension.rb
+++ b/lib/scout_apm/extension.rb
@@ -9,7 +9,7 @@ module ScoutApm
       end
     else
       def extensions_module_for(klass)
-        klass.include(Module.new)
+        klass.send(:include, Module.new)
       end
       
       def apply(klass)
@@ -21,11 +21,11 @@ module ScoutApm
           
           parent = extensions_module_for(klass)
           
-          parent.define_method(name) do |*args, &block|
+          parent.send(:define_method, name) do |*args, &block|
             original_method.bind(self).call(*args, &block)
           end
           
-          klass.define_method(name, wrapper_method)
+          klass.send(:define_method, name, wrapper_method)
           
           return klass
         end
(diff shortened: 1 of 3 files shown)


ioquatix:
Please bear with me while I get these tests passing on 1.x - I can't get it running locally (doesn't even compile) so I need to do some work here using travis.

ioquatix:
I found out that `super` is early bound in 1.x and late bound in 2.x so it was never going to work. That being said, it IS possible if we just use a block instead of supplying a module, as we can use the same technique. We move all the methods in the class up to an anonymous module, and `class_eval` replacements. I got 1.9 working locally, and it seems to be a winner.

ioquatix:
Damn, it worked on 1.9.3 but not 1.8.7 with a super cryptic error message:

> superclass method `log` disabled

ioquatix:
Okay, this approach works on 1.8.7, 1.9.3, and 2.0+, using prepend where possible.

It means you need to write extensions like this:

```
ScoutApm::Extension.apply(ThingYouWantToExtend) do
  def method_to_wrap
    super # invoke wrapped implementation
  end
end
```

It gives a unified method for all versions of Ruby, and can fix the issues we have with AR infinite recursion since it will use `Module#prepend` in that situation.

ioquatix:
As discussed, we probably don't take this approach.

cschneid:
But, @matthewhazlehurst this would be a good short, highly technical blog post about some of the weirder edges of ruby.

ioquatix:
I can write a blog post about it :p

matthewhazlehurst:
@ioquatix Awesome, that would be great!

PR #270, opened 2019-07-04
Specify minimum Ruby version "~> 2.1"

ioquatix:
As discussed, it might make sense to set the minimum version for Ruby in the 3.x release of Scout. This PR shows how we can do it, updates part of the ActiveRecord instrumentation to use `Module#prepend`.

Builds on https://github.com/scoutapp/scout_apm_ruby/pull/271

Issue #217, opened 2018-09-06
Use with Octoshark gem `stack level too deep` exception


drl123:
Trying to replace NewRelic with Scout in our Rails 5.0 app (ActiveRecord v5.0.3).  We use Octoshark to manage sharding of user data into multiple db's, and the main db uses the default AR connection.   I'm getting a `stack level too deep` exception when the instrumenter tries to run the query against the shards.  I've tried the 2.4.17, 2.4.11 and 3.0.0-pre28 (after looking through the issues about `prepend`)and they all do the same.  Any suggestions for how to configure these two gems to be compatible?  (I have pulled the NR gem completely out too, so I know it isn't a collision between instrumenters and NR instrumented these just fine with no config changes).

![screen shot 2018-09-06 at 1 04 39 pm](https://user-images.githubusercontent.com/568926/45173489-54e1c500-b1d6-11e8-824f-1729ce7ade1f.png)

cschneid:
@drl123 This [line](https://github.com/dalibor/octoshark/blob/master/lib/octoshark/active_record_extensions.rb#L24) in Octoshark is the issue.

Ruby's metaprogramming has evolved awkwardly over the years. Scout [overrides](https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/instruments/active_record.rb#L175) the ActiveRecord `log` method with a tracing version. To support older versions of Ruby, we use the `alias_method` approach to overriding.

Both approaches work in isolation, but the interaction of the two overriding the same method can cause loops like this.

We have some experimental code to attempt to support both (See #178 for an example), but haven't yet settled on this approach since it adds a fair bit of complexity.

drl123:
Thanks @cschneid...I did see that issue and a few others which is why I tried the various versions and didn't see any way to configure it for `prepend` vs `alias_method`.  Unfortunately, this is a show-stopper for us to make the switch to Scout as changing the sharding gem is a much more invasive endeavor.  If you do get resolution, please let me know.  We're eager to try out your platform.  I'd be happy to test any proposed solutions too.

Zeneixe:
We have the same issue, we're looking at using scoutapp but we can't part from octoshark. Any ETA for a fix?

dlanderson:
@ioquatix would be awesome to address this. We did some work in https://github.com/scoutapp/scout_apm_ruby/pull/178 last year but we need to review that progress and whether it's still the best option.

ioquatix:
[...]

PR #271, opened 2019-07-04
ActiveRecord quick fix

ioquatix:
This PR implement a quick fix and regression tests to sort out https://github.com/scoutapp/scout_apm_ruby/issues/217

I've updated the travis test matrix in order to expand our ability to test combinations of different gems using the `gems/*.gemfile` approach which I've used in the past, and implemented a very basic test case for ActiveRecord.

Finally, I discussed the best way to fix this with @cschneid and we decided that a quick fix was preferable, and the mid-term solution is to deprecate Ruby < 2.1 going forward, but that requires a bit more effort. So, this PR is good to be merged IMHO, and paves the way for 3.x.

See also https://github.com/scoutapp/scout_apm_ruby/pull/270

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ coverage/*
 lib/*.bundle
 lib/*.so
 log/scout_apm.log
+gems/*.lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,21 @@
 language: ruby
-
-rvm:
-  - "1.8.7"
-  - "1.9.3"
-  - "2.0"
-  - "2.2"
-  - "2.4"
-  - "2.5"
-
 cache: bundler
 
-before_install:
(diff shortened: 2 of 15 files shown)


ioquatix:
@itsderek23 can you review this?

ioquatix:
@cschneid should we merge this?

itsderek23:
Thanks for the ping @ioquatix ... I'm not too familar with this one.

I believe @cschneid is heads-down on our PHP agent and is pretty stretched for time right now ... I'll take a pass late next week.

dlanderson:
Sorry about that @ioquatix - I’ll review Monday. Thanks for the quick turnaround!

ioquatix:
I am going to merge this in the next day or two.

ioquatix:
Thanks for merging this!

PR #272, opened 2019-07-23
Ignores autoinstrument layers under 5ms in total call time and naming tweaks

itsderek23:
Autoinstrument layers are more likely to have fast execution times that are below a reasonable threshold for optimization. This discards autoinstrument layers that are under 5ms in total_call_time.


itsderek23:
This also includes some changes to we create layers:

* Using the code expression as the metric name
* Setting a partial backtrace, which we're able to collect from the source file path.

How it looks in the UI:

image

itsderek23:
Tests are broken because of the naming layer naming change. Investigating.

ioquatix:
Do you want me to merge this into auto_instruments branch?

itsderek23:
Yes

ioquatix:
Okay it's merged. Let's work on `auto_instruments` branch now if it makes sense.

ioquatix:
It's merged.

Issue #148, opened 2018-01-11
DevTrace triggers failed XHR requests w/large response headers


sdemjanenko:
I found that XHRs were returning 502 response codes because the DevTrace middleware was attaching a large header which was overwhelming my development server's nginx buffer. I increased the buffer size and the error went away.

itsderek23:
Note that this may appear in Chrome DevTools Console w/the error message "Failed to load resource: net::ERR_RESPONSE_HEADERS_TOO_BIG"

synth:
Hitting this as well when accessing our site via ngrok

ioquatix:
58KB in a single header!? As someone who implemented both HTTP/1, and HTTP/2, client & servers in both Ruby and C++, I'm having a 
![source](https://user-images.githubusercontent.com/30030/64473894-b610ef80-d1c0-11e9-8161-2819dc918ac4.gif)

Sorry, I didn't add anything to the issue. 58KB is a lot of data to put in a header.

synth:
Took a stab at an initial PR to add compression support in #282 ^

synth:
How come this was closed? I got the notification of this being closed and couldn't remember the initial context from 2019 and then separately was using DevTrace again and ran into this exact issue. Ended up finding this issue and my PR. I fetched upstream into my PR and it worked perfectly to address this issue. Guessing it was closed since no one else appears to encountering this. IMHO, header compression should still be considered :)

PR #282, opened 2019-09-07
Header compression support

synth:
Addresses #148 - Scout's trace header is too long. This occurs when trying to use ScoutAPM using a tool like ngrok which (currently) limits headers to 16KB. 

This PR is a rough pass on adding compression. In my tests it brings a 58KB header down to 8KB - which for me, gets ScoutAPM working with ngrok again. 

I think the solution in Ruby is pretty good by using Zlib for compression. However, the frontend decompression needs a lot of work. It currently inlines Pako (js zlib port) into a separate `xmlhtml_instrumentation.html` file. 

The long term solution will involve:
1. not inlining Pako
2. Only have one instrumentation file and have the config setting checked for compression checked

I don't have the time to take this further, but I'm putting this out there if anyone ones to take the baton.


dlanderson:
We'll have to pass on this PR for security purposes. The risk outweighs the benefit and haven't had many reports of the headers being an issue.

PR #289, opened 2019-10-01
Update LICENSE.md

### LINKED ISSUE #243: Relicense under MIT like Python agent
I run a service which uses AGPL code where I'm trying to debug some Rails code with memory issues. This means I have to provides the libraries for the service under the AGPL which means they must be under an AGPL compatible license. I can't use Scout because your Ruby library is under a proprietary license. I want to pay you money but your agent makes it impossible.

This is particularly annoying because your Python agent is under the MIT license. Is there any way you can relicense your Ruby agent under MIT as well?

matthewford:
closes #243 

Taken from the python agent.


wwahammy:
Hey, any chance this could be merged in?

PR #297, opened 2019-10-25
[WIP] Total trace time is sometimes incorrect

bohdanbenov:
For some reasons on v1 traces always one `SQL#other` is ignorable. I removed `skip_layer?` check and now on both versions of traces AR has the same `call_count` and response time.

Closes scoutapp/apm#628

### DIFF
--- a/lib/scout_apm/layer_converters/slow_request_converter.rb
+++ b/lib/scout_apm/layer_converters/slow_request_converter.rb
@@ -70,7 +70,6 @@ module ScoutApm
         allocation_metric_hash = Hash.new
 
         walker.on do |layer|
-          next if skip_layer?(layer)
           store_specific_metric(layer, metric_hash, allocation_metric_hash)
           store_aggregate_metric(layer, metric_hash, allocation_metric_hash)
         end


lib/scout_apm/layer_converters/slow_request_converter.rb
```diff
@@ -70,7 +70,6 @@ def create_metrics
         allocation_metric_hash = Hash.new
 
         walker.on do |layer|
-          next if skip_layer?(layer)
```
cschneid: `skip_layer` goes into why we skip some layers. I need more detail on what you found to trust that this fix is correct.

```
      # Sometimes we start capturing a layer without knowing if we really
      # want to make an entry for it.  See ActiveRecord instrumentation for
      # an example. We start capturing before we know if a query is cached
      # or not, and want to skip any cached queries.
      def skip_layer?(layer)
        return false if layer.annotations.nil?
        return true  if layer.annotations[:ignorable]
      end
```

The issue you linked is almost certainly a APM UI issue that can crop up in rare cases, not an agent data collection issue.

bohdanbenov: Yes, I agree it looks like the APM UI issue. I found out that we always have a different total response time on both versions of traces and also between the tree and summary section on the first type of traces. There are a few causes of it. We are using a `filtered_nodes` method in the _tree.html.slim partial. It rejects traces that have response time less than 1ms. It seems insignificant, but when we have a lot of these traces they are important. 
For example, this is a [condensed view](https://img.in6k.com/screens/e2ce906c_2019.28.10.png), and on the expanded trace view ([screenshot](https://img.in6k.com/screens/fd47514d_2019.28.10.png)), we can see all traces that have response time less than 1ms. After calculation:

- Condensed view: 441.5ms
- Expanded view: 447.97ms (the same as in the summary section)

So, I fixed it just added the [checkbox on the summary version of traces](https://github.com/scoutapp/apm/pull/1975), and the difference between the tree duration and total values on the summary section was eliminated. 
But, there is still the difference between both versions([tree](https://img.in6k.com/screens/3fbdbb63_2019.28.10.png) and [timelines](https://img.in6k.com/screens/a4acc623_2019.28.10.png)) of traces (see response time numbers and Active Record calls). I found out that this difference presents in all these places:

- on the controller part (endpoints_controller.rb ->@timing_tree and @timing_summary)
- on DB (request table and serialized_metrics column)
- at the point where we get the data from the agent ( checkin_controller#checkin)

After the research on the agent side, I figured out that `skip_layer?` condition in the `slow_request_converter.rb` file causing this difference. As far as I understood, our metrics data are creating in that part of the agent and after that, it sends to the APM side, writes to the database as `serialized_metrics`. So, in this place, one of SQL#other trace is always skipped, because it has annotation status ignorable. At the same time, the same SQL#other trace always present in the spans data, and accordingly on the timeline version of traces.
So, both types of traces will be exactly the same and the [problem described by Dave](https://github.com/scoutapp/apm/issues/628#issuecomment-314196089) also will be fixed after deleting that condition.

cschneid: That is an interesting theory. That could be the underlying mismatch - either counting AR cached SQL calls or not inconsistently. However, even if this is the root of a mismatch, it's not right to remove this line of code. It was added to have 'sql' instruments actually be real sql calls, not cached data. Later we may write a distinct instrument for cached calls, but not right now.

The issue you link back to on APM's repo predates any of our TraceV2 logic (individual spans), and both the SlowRequestConverter and the MetricsConverter classes call `skip_layer?` at the appropriate spot , so their view of the world matches up. 

You do bring up a good point to look at with the new TraceConverter, and double checking that it doesn't include layers that it shouldn't (ignorable and ignore_children ones), but that's outside the scope of the issue.

PR #310, opened 2019-11-19
Add additional logging to identify Marshal.dump issues

cschneid:
Some data is sneaking into the Remote::Message Marshal.dump call sometimes in a
customer's environment. We're not sure what data it is though, so this change
will add more logging when an object isn't dumpable for some reason.

Currently logs at INFO level. We may need to move to DEBUG later on, although the quantity of output is pretty small.

The log output will look like:
```
Found undumpable object: ["Root - NestedThingy", "@y - Array", "Index 2 - NestedThingy", "@y - NestedThingy", "@y - Hash"]
```


cschneid:
Superceeded by https://github.com/scoutapp/scout_apm_ruby/pull/312

PR #322, opened 2020-02-05
Fix depreciation warnings in Rails 6

emgreen33:
Calling ActionController::Module causes depreciation warnings as a result
of autoloading.
These get autoloaded by Rails when we initialize the app which
causes Rails 6+ to raise depreciation warnings.

### DIFF
--- a/lib/scout_apm/instruments/action_controller_rails_3_rails4.rb
+++ b/lib/scout_apm/instruments/action_controller_rails_3_rails4.rb
@@ -22,10 +22,10 @@ module ScoutApm
         # before and after filter timing. Instrumenting Base includes those
         # filters, at the expense of missing out on controllers that don't use
         # the full Rails stack.
-        if defined?(::ActionController)
+        ActiveSupport.on_load(:action_controller) do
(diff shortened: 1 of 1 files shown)


cschneid:
I'm going to be taking a look at these deprecation warnings, but we need to maintain compatibility with non-rails web apps too during startup. We support plain Rack apps, and Sinatra where `ActiveSupport` may not exist, causing an error at startup.

There's probably a clean answer for this, perhaps simply catching the `NameError` that would be emitted.

cschneid:
I reimplemented this in another PR (#337) with additional checks for it to be safer for non-rails environments, and to work with the weird scoping of `self` in the on_load block..

Thank you for the patch, the fix will be going out with our next release soon.

emgreen33:
> I reimplemented this in another PR (#337) with additional checks for it to be safer for non-rails environments, and to work with the weird scoping of `self` in the on_load block..
> 
> Thank you for the patch, the fix will be going out with our next release soon.

Great! Thanks for updating it, and the detailed explanation.

PR #324, opened 2020-03-23
Fix typo when calling `queue_time_ms`

tcannonfodder:
* In a TrackedRequest, the `queue_time` is present, but the `queue` is
	`nil` (because we don't know what the queue is, just that we've been
	given a queue time via a request header)


cschneid:
Thank you!

PR #334, opened 2020-05-18
Delete .DS_Store

dlackty:


cschneid:
Stupid sneaky ds_store! It's been lurking there since 2016!

Issue #333, opened 2020-05-08
Allow user defined `scout_transaction_uri` in ActionControllers


dlanderson:
Use case: A user's Rail's app has some routes where the route parameter might not be something they want to record in a detailed trace. E.g.

`/users/:email/edit` would look like `/users/foo@example.com/edit` as gathered in [scout_transaction_uri](https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/instruments/action_controller_rails_3_rails4.rb#L100).

* Add a `respond_to?(:scout_transaction_uri)` in the instruments so we can allow Scout users to define their own `scout_transaction_uri` and determine how the URI is recorded.
* Add a sanity check for what we get back from `scout_transaction_uri` - e.g. expect a non-blank string at minimum.

PR #341, opened 2020-06-26
User override of captured URI on Controllers

cschneid:
Scout captures the incoming URI and stores it for viewing on traces. There
are 2 pre-existing built in security modes, one that captures the whole
thing, and one that trims off any query params (?foo=bar)

But some users have more advanced needs for manipulating the URI before
capturing. Typically these involve times where personally identifiable
info gets put into the URL itself. The motivating example of this is
a url that looks something like /users/chris@scoutapm.com/dashboard.

The customer with that URL doesn't want to send all of their customer's
email addresses up, but also doesn't want to give up uri capturing
generally.

The new API is to override the `scout_transaction_uri` function in
the controller, returning a string with the modified uri. To get the
built in version, call `super`.

For instance:

```ruby
def scout_transaction_uri
  if action_name == "show"
    request.path.gsub(/\d+/, ":id")
  else
    super
  end
end
```

### DIFF
--- a/lib/scout_apm/instruments/action_controller_rails_3_rails4.rb
+++ b/lib/scout_apm/instruments/action_controller_rails_3_rails4.rb
@@ -74,8 +74,7 @@ module ScoutApm
       # before_action callbacks
       def self.build_instrument_module
         Module.new do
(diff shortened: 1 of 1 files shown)


cschneid:
Fixes #333

PR #347, opened 2020-07-22
Error service reporter

cschneid:
This PR is for preliminary support of error reporting. Contact support@scoutapm.com if you're interested in being part of our beta program.

Captures exceptions via a middleware and reports them up to Scout.


lib/scout_apm/middleware.rb
```diff
@@ -26,7 +26,8 @@ def attempt_to_start_agent
       ScoutApm::Agent.instance.start
       @started = ScoutApm::Agent.instance.context.started? && ScoutApm::Agent.instance.background_worker_running?
     rescue => e
-      ScoutApm::Agent.instance.context.logger("Failed to start via Middleware: #{e.message}\n\t#{e.backtrace.join("\n\t")}")
+      binding.pry
```
dlanderson, line 29: 🚨 👮

cschneid: Good catch

Issue #319, opened 2020-01-24
Ruby 2.7 Deprecation Warning


oniofchaos:
```
/app/vendor/bundle/ruby/2.7.0/gems/scout_apm-2.6.6/lib/scout_apm/instruments/active_record.rb:333: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

ioquatix:
@cschneid would you like me to sort this out?

dlanderson:
Thanks! This was addressed in https://github.com/scoutapp/scout_apm_ruby/pull/354 and released in 4.0.0 - Closing

PR #354, opened 2020-09-13
Fix a handful of Ruby 2.7 deprecation warnings

jasonrudolph:
This pull request resolves three deprecation warnings seen when using the latest version of scout_apm (v2.6.9) with Ruby 2.7.1.

4c7ac9f94fd99b1e94fdeb1b09b41d7099b9e19c resolves the specific warning cited in https://github.com/scoutapp/scout_apm_ruby/issues/319. This warning is seen when using `ActiveRecord::Querying.find_by_sql`:

    gems/scout_apm-2.6.9/lib/scout_apm/instruments/active_record.rb:333: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
    gems/activerecord-6.0.3.3/lib/active_record/querying.rb:45: warning: The called method `find_by_sql_without_scout_instruments' is defined here

1aa5f92e1a787da122167fcba425c44f59ebe440 resolves the following warnings seen using `ActiveRecord::Persistence#save` and `ActiveRecord::Persistence#save!`:

    gems/scout_apm-2.6.9/lib/scout_apm/instruments/active_record.rb:421: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
    gems/activerecord-6.0.3.3/lib/active_record/suppressor.rb:43: warning: The called method `save' is defined here

    gems/scout_apm-2.6.9/lib/scout_apm/instruments/active_record.rb:437: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
    gems/activerecord-6.0.3.3/lib/active_record/suppressor.rb:47: warning: The called method `save!' is defined here

In all of these cases, the ActiveRecord method accepts a combination of positional args and keyword args. We can resolve these warnings by updating the method signatures in scout_apm to 1) accept the same combination of positional args and keyword args, and 2) pass along those positional args and keyword args when invoking the corresponding method in ActiveRecord.

### DIFF
--- a/lib/scout_apm/instruments/active_record.rb
+++ b/lib/scout_apm/instruments/active_record.rb
@@ -408,7 +408,7 @@ module ScoutApm
     end
 
     module ActiveRecordUpdateInstruments
-      def save(*args, &block)
+      def save(*args, **options, &block)
         model = self.class.name
         operation = self.persisted? ? "Update" : "Create"
 
@@ -418,14 +418,14 @@ module ScoutApm
         req.start_layer(layer)
         req.ignore_children!
         begin
-          super(*args, &block)
+          super(*args, **options, &block)
         ensure
           req.acknowledge_children!
           req.stop_layer
         end
       end
 
-      def save!(*args, &block)
+      def save!(*args, **options, &block)
         model = self.class.name
         operation = self.persisted? ? "Update" : "Create"
 
@@ -434,7 +434,7 @@ module ScoutApm
         req.start_layer(layer)
         req.ignore_children!
         begin
-          super(*args, &block)
+          super(*args, **options, &block)
         ensure
           req.acknowledge_children!
           req.stop_layer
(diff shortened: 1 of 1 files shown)


cschneid:
Thank you for the changes @jasonrudolph! We really appreciate it.

There is a wrinkle with our (probably overzealous) support for very old versions of Ruby: the `**kwargs` does not parse on Ruby 1.8.7. We do still have customers on that old of a Ruby version, so we need a proper approach to handling the pressure for the breaking changes to make the agent better for new Rubies.

Give me a bit to figure out the best approach and I'll get this merged along with that. I'll merge this as soon as I can.

```
irb(main):002:0> def foo(*args, **kwargs, &block)
irb(main):003:1> puts args.inspect
irb(main):004:1> puts kwargs.inspect
irb(main):005:1> puts block.call()
irb(main):006:1> end
SyntaxError: compile error
(irb):2: syntax error, unexpected tPOW, expecting tAMPER or '&'
def foo(*args, **kwargs, &block)
                 ^
	from (irb):2
	from :0
irb(main):007:0> RUBY_VERSION
=> "1.8.7"
```

PR #360, opened 2020-10-16
Fix escaping real backslashes in payload json

cschneid:
Needs some more testing. This should fix a reported issue with escaping real backslashes in the payload. In the reported case, it was in the middle of some SQL, where the customer was storing escaped JSON as a string - the escaping we had previously failed to generate valid JSON.

Note, this entire json serializer should be removed, and we can begin relying on Ruby's built-in JSON library once we bump the minimum version to Ruby 2.1+. We only have it to avoid an extra dependency in Ruby 1.8.7.

### DIFF
--- a/lib/scout_apm/serializers/payload_serializer_to_json.rb
+++ b/lib/scout_apm/serializers/payload_serializer_to_json.rb
@@ -45,6 +45,7 @@ module ScoutApm
           "{#{str_parts.join(",")}}"
         end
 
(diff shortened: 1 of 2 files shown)


cschneid:
This is now failing tests on Ruby 1.8.7, with its escaping. Did gsub change significantly? Or maybe the regex engine change back then is doing it? 

Don't merge yet, needs to be fixed.

PR #365, opened 2020-12-19
Resolve Rails 6.1 deprecation warning re: connection_config

jasonrudolph:
Prior to this change, the first interaction with the database would emit the following deprecation warning with Rails 6.1:

> DEPRECATION WARNING: connection_config is deprecated and will be removed from Rails 6.2 (Use connection_db_config instead)

The deprecation was introduced in https://github.com/rails/rails/pull/38005 and includes the following [release notes](https://github.com/rails/rails/pull/38005/files#diff-353f4f101b1bb04938d761b6e29c88e0f147084e9e276854a77ae5d9a9a3b2a8):

> The `connection_config` method has been deprecated, please use `connection_db_config` instead which will return a `DatabaseConfigurations::DatabaseConfig` instead of a `Hash`.

### Proposed resolution

With the changes in the pull request, Scout first attempts to use `connection_db_config`. If unsuccessful, Scout falls back to the previous behavior of using `connection_config`.

### How to reproduce the deprecation warning

1. Verify you're using Rails 6.1:

    ```
    $ rails -v
    Rails 6.1.0
    ```

2. Set up a new Rails app with scout_apm:

    ```
    $ rails new sample --database=postgresql
    $ cd sample/
    $ rails db:create
    $ rails generate model User name:string
    $ rails db:migrate
    $ bundle add scout_apm 
    ```

3. Load Scout and interact with the database:

    ```
    $ SCOUT_DEV_TRACE=true rails console

    2.7.2-p137> User.count
    DEPRECATION WARNING: connection_config is deprecated and will be removed from Rails 6.2 (Use connection_db_config instead) (called from irb_binding at (irb):1)
      (0.4ms)  SELECT COUNT(*) FROM "users"
    => 0
    ```

### DIFF
--- a/lib/scout_apm/framework_integrations/rails_3_or_4.rb
+++ b/lib/scout_apm/framework_integrations/rails_3_or_4.rb
@@ -71,7 +71,11 @@ module ScoutApm
       #
       # We avoid this issue by not calling .respond_to? here, and instead using the less optimal `rescue nil` approach
       def raw_database_adapter
-        adapter = ActiveRecord::Base.connection_config[:adapter].to_s rescue nil
+        adapter = ActiveRecord::Base.connection_db_config.configuration_hash[:adapter] rescue nil
+
+        if adapter.nil?
+          adapter = ActiveRecord::Base.connection_config[:adapter].to_s rescue nil
+        end
 
         if adapter.nil?
           adapter = ActiveRecord::Base.configurations[env]["adapter"]


cschneid:
Thank you for this PR. Sorry it's taking a bit to take a look at, lots of time off over holidays 😄 .  I'll get this tested & merged as soon as I can.

lib/scout_apm/framework_integrations/rails_3_or_4.rb
```diff
@@ -71,7 +71,11 @@ def database_engine
...
       def raw_database_adapter
-        adapter = ActiveRecord::Base.connection_config[:adapter].to_s rescue nil
+        adapter = ActiveRecord::Base.connection_db_config.configuration_hash[:adapter] rescue nil
+
+        if adapter.nil?
+          adapter = ActiveRecord::Base.connection_config[:adapter].to_s rescue nil
+        end
 
         if adapter.nil?
           adapter = ActiveRecord::Base.configurations[env]["adapter"]
```
dieggofaustino, line 81: Here is another depreciation, using `[]` for the `configurations` method. If it is your intention please consider replacing it with `configs_for`
```ruby
adapter = ActiveRecord::Base.configurations.configs_for(env)["adapter"]
```
https://github.com/rails/rails/pull/33637

dieggofaustino, line 81: Here is another depreciation, using `[]` for the `configurations` method. If it is your intention please consider replacing it with `configs_for`
```ruby
adapter = ActiveRecord::Base.configurations.configs_for(env)["adapter"]
```
https://github.com/rails/rails/pull/33637

cschneid: We have to be backwards compatible a long ways, so we can't just change it to that blindly. Do we hit this branch on 6.1+ after checking the line above with `connection_db_config`? 

I'll have to check it out

ddarren:
> Thank you for this PR. Sorry it's taking a bit to take a look at, lots of time off over holidays 😄 . I'll get this tested & merged as soon as I can.

@cschneid Any news on getting this merged or when we can expect it? Still getting this on 4.0.4. Thanks!

rubendinho:
Hi there - are there any updates on this?

benhart:
Hi - any idea when this will be merged and released? Or is there another way for us to avoid the deprecation warnings?

jasonrudolph:
Thanks for getting this merged in, @cschneid. 🙇

PR #371, opened 2021-01-13
typhoeus support

mjobin-mdsol:
Unsure if this is complete yet... 

I am testing it, I never tried to add instrumentation to this gem.
And I am not at-all familiar with Typhoeus internals.

So, any help appreciated here.

Thanks


cschneid:
Thanks for the PR, I'll take a look at Typhoeus. Its sorta tricky to find the right "chokepoint" method in libraries to instrument.

I'll ping you here as soon as I add it.

cschneid:
Implemented in #376. Thank you for this PR!

PR #376, opened 2021-01-19
Add Typhoeus instrument for both single and hydra versions

cschneid:
Thank you @mjobin-mdsol for #371, rewrote much of it - Typhoeus is sorta a tricky library. I think I've got it nailed down though for basic and hydra operation. 

When we get some time to look into this, and if there's demand, we can do more to breakdown the calls inside of Hydra batches. But for now we just report the overall time and count.


cschneid:
Released in 4.0.2

Issue #377, opened 2021-01-20
ArgumentError in render after upgrading to 4.0.2


adrianpacala:
Hey everyone.

We ran into an issue after upgrading to Scout APM 4.0.2. We're suspecting it has something to do with changes to signatures of rendering methods introduced in 29396793c91efef74e7f210951f07d0d94ba0bcb.

For example, this call to `render` which used to be working fine with 4.0.0:

```ruby
render 'home', layout: 'home'
```

now raises an error:

```ruby
ArgumentError - wrong number of arguments (given 3, expected 4):
  app/controllers/home_controller.rb:8:in `show'
```

We found similar issues in other gems:

- activeadmin/activeadmin/issues/5821
- activerecord-hackery/ransack/issues/948

The versions we're using are:

```
[1] pry(main)> RUBY_VERSION
=> "2.6.1"
[2] pry(main)> Rails.version
=> "6.0.3.4"
[3] pry(main)> ScoutApm::VERSION
=> "4.0.2"
```

`scout_apm.yml` below:

```
# This configuration file is used for Scout APM.
# Environment variables can also be used to configure Scout. See our help docs at http://help.apm.scoutapp.com#environment-variables for more information.
common: &defaults

  # key: Your Organization key for Scout APM. Found on the settings screen.
  # - Default: none
  key: <%= ENV['SCOUT_APM_KEY'] %>

  # log_level: Verboseness of logs.
  # - Default: 'info'
  # - Valid Options: debug, info, warn, error
  # log_level: debug

  # name: Application name in APM Web UI
  # - Default: the application names comes from the Rails or Sinatra class name
  name: xxx

  # monitor: Enable Scout APM or not
  # - Default: none
  # - Valid Options: true, false
  monitor: true

production:
  <<: *defaults

development:
  <<: *defaults
  monitor: false

test:
  <<: *defaults
  monitor: false

staging:
  <<: *defaults
```

cschneid:
I think I've got this fixed in 4.0.3, can you give it a try?

benedictfritz:
Ran into this same error. Updating to 4.0.3 fixed it.

adrianpacala:
Can confirm updating to 4.0.3 fixed this issue for us.

schneikai:
[...]

PR #379, opened 2021-01-20
Pass kwargs to render_partial instrument in TemplateRenderer (for locals)

cschneid:
But don't do that in the PartialRenderer instrument, which has different interface that has locals as an instance var, not passed to the method. (Rails 6.0.3, Rails 4.2)


cschneid:
Fixes #377

PR #381, opened 2021-01-26
Fix keyword args warning in instrument_method

betesh:
This is a follow-up to #354


cschneid:
Thank you for the fix! This will be out in the next release, next week.

PR #384, opened 2021-02-04
Removing Hack for Regexp supporting ruby 1.8.7

esparta:
This hack, added on Oct 2020 (#360) was a way to fix a use case related to Regexp in ruby 1.8.7

Since Scout APM 4.0.0 (Nov 2020), the gem only support ruby ~> 2.1 (#270).

Is not longer possible to use ruby 1.8.7, the hack can be safely removed.

https://github.com/scoutapp/scout_apm_ruby/pull/360
https://github.com/scoutapp/scout_apm_ruby/pull/270

### DIFF
--- a/lib/scout_apm/serializers/payload_serializer_to_json.rb
+++ b/lib/scout_apm/serializers/payload_serializer_to_json.rb
@@ -45,31 +45,17 @@ module ScoutApm
           "{#{str_parts.join(",")}}"
         end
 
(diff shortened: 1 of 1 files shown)


cschneid:
Thank you for this - we have a fair bit of legacy code in the agent (this being one bit) to deal with 1.8.7. Expect to see more of this kind of cleanup soon as I get time too.

Issue #391, opened 2021-02-16
Active record instrumentation not expecting current_layer.desc to be a string


shadwell:
I'm getting an error in instruments/active_record#log where current_layer.desc is a String.

```
NoMethodError: undefined method `merge' for "xxx":String
    scout_apm (4.0.4) lib/scout_apm/instruments/active_record.rb:255:in `log'
    activerecord (6.1.1) lib/active_record/connection_adapters/abstract_mysql_adapter.rb:204:in `execute'
    activerecord (6.1.1) lib/active_record/connection_adapters/mysql/database_statements.rb:50:in `execute'
```

It looks like it might be the Typhoeus instrumentation setting the desc to be the URL it's dealing with and then the active_record instrumentation expecting current_layer.desc to be an `SqlList`.

What I'm doing is calling Typhoeus in a callback in the active record model (in order to index a saved record to elasticsearch).

dennmart:
I'm also receiving this error when an exception is captured in our app's error tracker (Sentry). I thought it was similar to what was occurring in #378 but that has already been fixed.

wilg:
Getting this too

wilg:
@cschneid Any idea what this might be? It's preventing me from using Scout.

dennmart:
@wilg I downgraded the scout_apm gem to 4.0.1 (before the library added Typhoeus instrumentation, which seems to be creating this issue) and it's working well. You can temporarily use that version as a workaround.

wilg:
Great, thanks!

lomefin:
We had this issue as well.
typhoeus.rb is returning desc as a string in other instrument files the layer is a Hash

PR #392, opened 2021-02-22
Set up desc on new layers before starting the layer

shadwell:
Set up desc on new layers before starting the layer rather than setting it on the current layer after start_layer.

To fix : https://github.com/scoutapp/scout_apm_ruby/issues/391

This means that if the current layer isn't the one we just tried to start then the desc isn't set on a layer we weren't expecting.

I haven't managed to reproduce this in a test (it's not clear to me the best way to test layering integration of different instrumentation).

Reproducing in a separate application was achieved by an active record callback calling Typhoeus (through elasticsearch) and then further active record operations taking place.

### DIFF
--- a/lib/scout_apm/instruments/typhoeus.rb
+++ b/lib/scout_apm/instruments/typhoeus.rb
@@ -29,10 +29,11 @@ module ScoutApm
 
       module TyphoeusHydraInstrumentation
         def run(*args, &block)
+          layer = ScoutApm::Layer.new("HTTP", "Hydra")
+          layer.desc = scout_desc if current_layer
+
           req = ScoutApm::RequestManager.lookup
-          req.start_layer(ScoutApm::Layer.new("HTTP", "Hydra"))
-          current_layer = req.current_layer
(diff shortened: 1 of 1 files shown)


cschneid:
Thank you for this PR - I started in on the Typheous desc issue and came to the same conclusion as the first code change needed. This fix is obviously better, but I'll be doing more testing trying to characterize it better, since it appears that the underlying cause is that Typhoeus uses threads, or perhaps some other ordering issue. Starting a layer and then immediately asking for current layer shouldn't be so fraught.

delphaber:
Hi :)

I think this is breaking compatibility with older versions of Typhoeus gem (`1.3.1`).

Since the upgrade of `scout_apm` to version `4.1.0` we are experiencing this error:
> NameError: undefined local variable or method `current_layer' for #<Typhoeus::Hydra:0x00007f8f373d9118>

I've reverted `scout_apm` from `4.1.0` to `4.0.4` and I'll update our typhoeus gem version, but maybe you can consider the idea to check if `current_layer` method exists before calling it. What do you think? :)

Issue #393, opened 2021-02-23
Scout broken on Rails Master due to change on `log` in `abstract_adapter`


brenogazzola:
Hello there. We are using Scout and Rails Master on production and just stumbled into a problem caused by, I believe, Rails PR 40037.

This PR added the `load_async` methods in order to allow multiple queries to run in the background at the same time. It also changed the signature of the `log` method in the `abstract_adpater` which Scout seems to rely on. The error message we are seeing is `wrong number of arguments (given 6, expected 1..5)`. Below I have added the full stack trace. 

The cause seems to be the file `scout_apm/instruments/active_record.rb` which calls rails `log` (which was modified) on line 256.

Disabling Scout solved the problem for us.

```
irb(main):001:1* begin
irb(main):002:1*   Website.last
irb(main):003:1* rescue => e
irb(main):004:1*   puts e.backtrace
irb(main):005:1* end
/home/app/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/bundler/gems/rails-ab8e3d22cbac/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb:697:in `log'
/home/app/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/scout_apm-4.0.4/lib/scout_apm/instruments/active_record.rb:256:in `log'
/home/app/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/bundler/gems/rails-ab8e3d22cbac/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb:688:in `exec_cache'
/home/app/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/bundler/gems/rails-ab8e3d22cbac/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb:654:in `execute_and_clear'
/home/app/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/bundler/gems/rails-ab8e3d22cbac/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb:51:in `exec_query'
/home/app/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/bundler/gems/rails-ab8e3d22cbac/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb:547:in `select'
[...]

PR #394, opened 2021-02-23
Improve blind delegation of ActiveRecordInstruments#log

casperisfine:
Should fix: https://github.com/scoutapp/scout_apm_ruby/issues/393

### Context

Rails edge changed the signature of `#log`, it now includes a keyword arg (see https://github.com/rails/rails/pull/40037)

The only proper way to blind delegate keyword args across versions is to use ruby2_keywords, see https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html for details.

### DIFF
--- a/lib/scout_apm/instruments/active_record.rb
+++ b/lib/scout_apm/instruments/active_record.rb
@@ -215,6 +215,7 @@ module ScoutApm
           end
         end
       end
+      ruby2_keywords :log if respond_to?(:ruby2_keywords, true)
     end
 
     module ActiveRecordInstruments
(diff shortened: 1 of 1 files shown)


brenogazzola:
Any chance of merging this? We've disabled scout on production because `load_async` is more valuable for us than having Scout (at the moment, where we know the app is stable performance wise).

dlanderson:
@brenogazzola We'll get this released this week.  @casperisfine Thanks for the PR - much appreciated.

cschneid:
Thank you for the PR, it'll go out in our next release in a few days.

PR #402, opened 2021-04-30
prepend Net::HTTP instrumentation

flyerhzm:
It gets `stack level too deep (SystemStackError)` with the latest airbrake, which uses `prepend` to hack `Net::HTTP` since 10.0.2.

To fix the SystemStackError, we should also use `prepend` to instrument `Net::HTTP`

### DIFF
--- a/lib/scout_apm/instruments/net_http.rb
+++ b/lib/scout_apm/instruments/net_http.rb
@@ -22,33 +22,31 @@ module ScoutApm
 
           logger.info "Instrumenting Net::HTTP"
 
-          ::Net::HTTP.class_eval do
-            include ScoutApm::Tracer
+          ::Net::HTTP.send(:include, ScoutApm::Tracer)
+          ::Net::HTTP.send(:prepend, NetHttpInstrumentation)
+        end
+      end
 
-            def request_with_scout_instruments(*args, &block)
-              self.class.instrument("HTTP", "request", :ignore_children => true, :desc => request_scout_description(args.first)) do
-                request_without_scout_instruments(*args, &block)
-              end
-            end
-
-            def request_scout_description(req)
-              path = req.path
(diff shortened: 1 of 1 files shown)


st0012:
Since this SDK has dropped the support of Ruby 1.9, I think using `prepend` is better and I'd love to see this be merged as well 🙂

FraDim:
@cschneid, @dlanderson: This issue affects new versions of Airbrake and Sentry. What do you think of this approach?

cschneid:
I am prepping a 4.1.0 version of the agent with several bug fixes, and then after that looking at 5.0. The big breaking change is switching to all prepend by default, and adding a configuration option to use alias if that causes compatibility issues with other gems.

While every ruby we support also has prepend, many other gems use alias_method, and that causes issues when mixed. Keep an eye out for this change, should be up pretty quick.

mvastola:
Just wanted to check on the status of this. I know the expected timeline was stated as "pretty quick" in May, and it's been over 2 months. Is there an ETA?

lostie:
@itsderek23 @dlanderson @cschneid do you have an ETA for this change to be released?

davidalejandroaguilar:
Friendly reminder about this 😃

andychongyz:
Any update on when this will get merged?

pawurb:
Any chance to get this one out? We need it on prod.

dlanderson:
@flyerhzm Thank you very much for this PR. I'm sorry for the (way too long) delay. Please see https://github.com/scoutapp/scout_apm_ruby/pull/448 for how we'll be implementing `prepend`. All contributions very much appreciated!

dlanderson:
Working on https://github.com/scoutapp/scout_apm_ruby/pull/448 to get it released this week.

Issue #303, opened 2019-10-31
Exception with specific versions of Parser - appears to be Parser bug


cschneid:
A customer ran into an exception when using AutoInstruments, with Parser 2.6.3.0, where it errors out during loading with: 

```
  bundle exec rake assets:precompile
/usr/local/bundle/gems/builder-3.1.4/lib/builder/xchar.rb:111: warning: constant ::Fixnum is deprecated
warning: parser/current is loading parser/ruby25, which recognizes
warning: 2.5.5-compliant syntax, but you are running 2.5.3.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
rake aborted!
NameError: uninitialized constant Parser::Base
/usr/local/bundle/gems/parser-2.6.3.0/lib/parser/ruby25.rb:14:in `<module:Parser>'
/usr/local/bundle/gems/parser-2.6.3.0/lib/parser/ruby25.rb:13:in `<top (required)>'
/usr/local/bundle/gems/parser-2.6.3.0/lib/parser/current.rb:65:in `<module:Parser>'
/usr/local/bundle/gems/parser-2.6.3.0/lib/parser/current.rb:3:in `<top (required)>'
/usr/local/bundle/gems/scout_apm-2.6.2/lib/scout_apm/auto_instrument/parser.rb:2:in `<top (required)>'
/usr/local/bundle/gems/scout_apm-2.6.2/lib/scout_apm/auto_instrument/rails.rb:3:in `<top (required)>'
/usr/local/bundle/gems/scout_apm-2.6.2/lib/scout_apm/auto_instrument/instruction_sequence.rb:2:in `<top (required)>'
/usr/local/bundle/gems/scout_apm-2.6.2/lib/scout_apm/auto_instrument.rb:2:in `<top (required)>'
/usr/local/bundle/gems/scout_apm-2.6.2/lib/scout_apm.rb:201:in `block in <class:Railtie>'
/usr/local/bundle/gems/railties-4.2.11.1/lib/rails/initializable.rb:30:in `instance_exec'
```

Updating to Parser version 2.6.5.0, or downgrading to 2.6.2.x appears to fix this issue.

cschneid:
closing, as this appears to be a parser bug, isolated to only an already outdated version or two. If you run into this, update the Parser gem, and that should fix it. Please let us know if this reappears.

dira:
The context was:

[...]

PR #424, opened 2021-09-03
Set dependency version to parser

kentarohorie:
I found parser dependency. With parser < 2.5.0.0, app can't start.
I think parser's version should be specified and create PR.

This error is caused when app start.
```
/app/vendor/bundle/ruby/2.6.0/gems/scout_apm-4.1.2/lib/scout_apm/auto_instrument/parser.rb:3:in `<top (required)>': Parser::TreeRewriter was not defined (LoadError)
```

Parser::TreeRewriter was implemented in parser 2.5.0.0
- https://github.com/whitequark/parser/commit/b0a7695f25de1ff717d4e1c23f45e647ccbcdc54#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR16

### DIFF
--- a/scout_apm.gemspec
+++ b/scout_apm.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rake-compiler"
   s.add_development_dependency "addressable"
   s.add_development_dependency "activesupport"
-  s.add_runtime_dependency "parser"
+  s.add_runtime_dependency "parser", ">= 2.5.0.0"
 
   # These are general development dependencies which are used in instrumentation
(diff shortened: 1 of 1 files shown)


kentarohorie:
The version should be specified to >= 2.6.5..? 🤔 
I found this issue.
https://github.com/scoutapp/scout_apm_ruby/issues/303

dlanderson:
Thanks! I've decided we should fail gracefully and log instead of defining a version dependency. Closing this in favor of https://github.com/scoutapp/scout_apm_ruby/pull/440

PR #427, opened 2021-09-10
Add magic comment frozen_string_literal: false to files where strings are modified

danielmorrison:
I was running scout in an environment where strings are frozen by default (using `RUBYOPT="--enable=frozen-string-literal"`) and Scout started raising exceptions. 

The magic comments allow it to run in this environment, without affecting a "typical" setup, or changing code.  

Before this PR, running tests using `RUBYOPT="--enable=frozen-string-literal rake` on `scout_apm_ruby` fails. After, it passes.


dlanderson:
Thank you!

Issue #420, opened 2021-06-11
HTTP.rb instrumentation doesn't work for 4.1.0


jrothrock:
We aren't capturing HTTP calls made from the HTTP.rb library. 

Closing the following in favor of this issue:
https://github.com/scoutapp/scout_apm_ruby/issues/258

ibrahima:
Hmm, just looking into this again since this is preventing the new External Services dashboard from showing all of our external services. It looks like this [file has not changed since it was merged](https://github.com/scoutapp/scout_apm_ruby/commits/master/lib/scout_apm/instruments/http.rb), and on our end there wasn't an update to HTTP.rb that could have broken things. So I am not really sure why this would have been broken... but it would be really nice if it would work!

dlanderson:
Apologies it took this long! Resolved by https://github.com/scoutapp/scout_apm_ruby/pull/433, pending release by next week!

ibrahima:
No worries, thank you Dave for fixing it! I'm kinda amused by that one because I could have sworn I saw it working at one point.. it'll be great to have this working soon!

PR #433, opened 2021-11-11
Actually install http.rb instrumentation as HTTP::Client

dlanderson:


dlanderson:
Resolves https://github.com/scoutapp/scout_apm_ruby/issues/420

Issue #421, opened 2021-06-30
"Failed loading configuration file" error with Psych 4


ybiquitous:
Hello, I encounter the following error with Psych 4 and Rails 6.1.4:

```console
$ bin/rails r 'puts Rails.version'
[Scout] [06/30/21 17:26:59 +0900 F200420NT01.local (380)] INFO : Failed loading configuration file (/tmp/repro/config/scout_apm.yml): Unknown alias: defaults. ScoutAPM will continue starting with configuration from ENV and defaults
6.1.4
```

Here is a part of `Gemfile`:

```ruby
gem 'scout_apm', '4.1.1'
gem 'psych', '4.0.1'
```

Here is `config/scout_apm.yml`:

```yaml
common: &defaults
  name: YOUR_APPLICATION_NAME
  key: YOUR_APPLICATION_KEY
  monitor: true

test:
  monitor: false

development:
  <<: *defaults

production:
  <<: *defaults
```

When using Psych 3, this error never occurs:

```diff
-gem 'psych', '4.0.1'
+gem 'psych', '< 4'
```

I think this is a similar problem with rails/rails#42257 (Psych 4 incompatibility).

dlanderson:
Resolved by #435 - pending release

ybiquitous:
@dlanderson Thank you!

PR #435, opened 2021-11-11
Allow using `YAML.unsafe_load` when reading scout_apm.yml

dlanderson:
Due to https://github.com/ruby/psych/pull/487 making YAML loads strict by default, when someone is running with Psych > 4 and Scout attempts to load `config/scout_apm.yml` with references, an exception is raised, caught, and logged, but the config is not loaded from the file.

Similar to the rails change https://github.com/rails/rails/pull/42257, we can consider configuration files as trusted.  This change reads the yml file with `unsafe_load` if available.


dlanderson:
Resolves https://github.com/scoutapp/scout_apm_ruby/issues/421

Issue #428, opened 2021-10-22
RuntimeError: can't add a new key into hash during iteration


mockdeep:
We see occasional errors coming from Scout in production. We're running version 4.1.2. Looks similar to [this previous issue](https://github.com/scoutapp/scout_apm_ruby/issues/205).

```
RuntimeError: can't add a new key into hash during iteration
 [GEM_ROOT]/gems/scout_apm-4.1.2/lib/scout_apm/request_histograms.rb:55 :in `block in initialize_histograms_hash`
 [GEM_ROOT]/gems/scout_apm-4.1.2/lib/scout_apm/request_histograms.rb:34 :in `add`
 [GEM_ROOT]/gems/scout_apm-4.1.2/lib/scout_apm/tracked_request.rb:192 :in `layer_insignificant?`
 [GEM_ROOT]/gems/scout_apm-4.1.2/lib/scout_apm/tracked_request.rb:118 :in `stop_layer`
 [GEM_ROOT]/gems/scout_apm-4.1.2/lib/scout_apm/auto_instrument/layer.rb:18 :in `AutoInstrument`
```

dlanderson:
Hi @mockdeep - would you mind giving [this branch](https://github.com/scoutapp/scout_apm_ruby/compare/v4_1_2_with_request_histogram_mutex) a try to see if it fixes your issue? In your gemfile:

`gem 'scout_apm', git: 'https://github.com/scoutapp/scout_apm_ruby.git', branch: 'v4_1_2_with_request_histogram_mutex'`

mockdeep:
Hey @dlanderson, thanks for the fix! Unfortunately, I don't think we're going to see a noticeable difference. That error happened once a month ago and we haven't seen it since. I'm not sure what we could do to properly test it.

dlanderson:
@mockdeep Was the `RuntimeError` just logged without otherwise affecting the app? If it's not easily reproduced and scarcely happens and does not affect the app, I think I would prefer to keep the potential fix as unreleased since adding the mutexes in a hot path can add overhead.

mockdeep:
@dlanderson it impacted a user who was trying to log into our platform. It's not a big deal, so if you want to close this I can report back in if we ever see it again.

dlanderson:
Waiting for feedback - closing for now.

PR #439, opened 2021-11-15
Add Mutex to RequestHistograms

dlanderson:
Closes https://github.com/scoutapp/scout_apm_ruby/issues/428


dlanderson:
Haven't had any other reports from https://github.com/scoutapp/scout_apm_ruby/issues/428 so closing for now.

Issue #404, opened 2021-05-04
Stack Level Too Deep errors w/ new sentry-ruby gem 4.4.0


navied:
Seems like a recent update to sentry gem is causing a bit of havoc with the scout gem see: https://github.com/getsentry/sentry-ruby/issues/1427

Would it be possible to do the required changes outlined above to make them behave properly?

Thanks!

PikachuEXE:
For those who needs a quick workaround (without downgrading scout gems)
But disables `net/http` instrument
```yaml
# config/scout_apm.yml
common: &defaults
  disabled_instruments:
    - "NetHttp"

development:
  <<: *defaults

test:
  <<: *defaults
  monitor: false

staging:
  <<: *defaults

production:
  <<: *defaults
```

st0012:
I've released `sentry-ruby 4.4.1`, which should solve the conflict issue now. Here's the related PR: https://github.com/getsentry/sentry-ruby/pull/1432
Sorry for the inconvenience.

mjobin-mdsol:
this issue is similar to #398 where legacy method aliasing conflicts with another gem using newer (2.0+) module#prepend

pjar:
We have the same issue but we don't use Sentry. It conflicts with NewRelic for us:


<img width="722" alt="Screenshot 2021-05-17 at 15 00 55" src="https://user-images.githubusercontent.com/236458/118492837-b9dc6f80-b720-11eb-8db2-d2547a72918c.png">

mjobin-mdsol:
@pjar have you tried update Scout to 4.x  ? or downgrading newrelic to 6.x or both

GregSmith92:
Same error with AppSignal 3.x

Does upgrading to scout_apm 4.x solve this issue?

danielmklein:
Hi there folks -- we are also seeing this with scout_apm 4.0.4 and newrelic_rpm 7.0.0. I tried some of the instrumentation config the New Relic team describes in the [7.0.0 upgrade guide](https://docs.newrelic.com/docs/agents/ruby-agent/getting-started/migration-7x-guide/#prepend-instrumentation) to no avail.

Downgrading newrelic_rpm to 6.15.0 resolved this for us, however.

eric-norcross:
I'm having this same issue using:

- ruby 2.6.6
- rails 6.0.4.4
- puma 4.3.10
- newrelic_rpm 8.3.0
- scout_apm 5.1.1

[...]

Issue #309, opened 2019-11-19
When using scout together with datadog, app can't start up


w32-blaster:
When we use scout_apm and ddtrace (datadog apm) as well we can't start up the application anymore. This was working before (version > 2.2). 
Here is the trace we get in rails:
"SystemStackError: stack level too deep"

dlanderson:
Apologies @w32-blaster - do you happen to have the versions of ddtrace and scout_apm you're using? If you also have a backtrace go ahead and place it here, too.

w32-blaster:
Sure. 
Versions we use:
scout_apm: latest released
ddtrace: 0.26.0

Here is the full stack trace:
```
SystemStackError: stack level too deep
  from active_support/core_ext/time/calculations.rb:285:in `minus_with_coercion'
  from ddtrace/contrib/rails/core_extensions.rb:80:in `render_template_with_datadog'
  from scout_apm/instruments/action_view.rb:128:in `render_template'
  from ddtrace/contrib/rails/core_extensions.rb:80:in `render_template_with_datadog'
  from scout_apm/instruments/action_view.rb:128:in `render_template'
  from ddtrace/contrib/rails/core_extensions.rb:80:in `render_template_with_datadog'
  from scout_apm/instruments/action_view.rb:128:in `render_template'
  from ddtrace/contrib/rails/core_extensions.rb:80:in `render_template_with_datadog'
  from scout_apm/instruments/action_view.rb:128:in `render_template'
  from ddtrace/contrib/rails/core_extensions.rb:80:in `render_template_with_datadog'
  from scout_apm/instruments/action_view.rb:128:in `render_template'
  from ddtrace/contrib/rails/core_extensions.rb:80:in `render_template_with_datadog'
  from scout_apm/instruments/action_view.rb:128:in `render_template'
  from ddtrace/contrib/rails/core_extensions.rb:80:in `render_template_with_datadog'
  from scout_apm/instruments/action_view.rb:128:in `render_template'
  from ddtrace/contrib/rails/core_extensions.rb:80:in `render_template_with_datadog'
  from scout_apm/instruments/action_view.rb:128:in `render_template'
  from ddtrace/contrib/rails/core_extensions.rb:80:in `render_template_with_datadog'
  from scout_apm/instruments/action_view.rb:128:in `render_template'
  from ddtrace/contrib/rails/core_extensions.rb:80:in `render_template_with_datadog'
[...]

PR #448, opened 2022-04-27
Initial commit for adding prepend in addition to alias_method

dlanderson:
For many of our instruments, we still use `Module#alias_method` to monkeypatch specific methods to gather telemetry. The more modern approach is to use `Module#prepend`, which many other gems now use as default.

Moving from `alias_method` to `prepend` has caused major headaches for nearly all gems which made the switch, despite making the change in a major version semantic version bump. In order to allow a less disruptive upgrade path, Scout will continue to use `alias_method` as the default, but with available configuration option to move either all instrumentation to `prepend`, or selectively move instrumentation to `prepend`.

We will consider making the `prepend` instrumentation default after this configurable option has been released for a sufficient amount of time.

This is the initial draft PR starting with the NetHTTP instrumentation (there's still some duplicate instrumentation code for NetHTTP). We'll need to do the same for all of the other instruments.

Resolves https://github.com/scoutapp/scout_apm_ruby/issues/404
Resolves https://github.com/scoutapp/scout_apm_ruby/issues/309
Resolves https://github.com/scoutapp/scout_apm_ruby/issues/114

Replaces PR https://github.com/scoutapp/scout_apm_ruby/pull/402

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,6 +40,13 @@ jobs:
           - ruby: 3.0
           - ruby: 3.0
             prepend: true
+          - ruby: 3.0
+            gemfile: gems/instruments.gemfile
+            test_features: "instruments"
+          - ruby: 3.0
+            gemfile: gems/instruments.gemfile
+            prepend: true
+            test_features: "instruments"
 
     env:
       BUNDLE_GEMFILE: ${{ matrix.gemfile }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,11 +35,16 @@ jobs:
             gemfile: gems/rails3.gemfile
             bundler: 1.17.3
           - ruby: 2.7
+          - ruby: 2.7
+            prepend: true
           - ruby: 3.0
+          - ruby: 3.0
+            prepend: true
 
     env:
       BUNDLE_GEMFILE: ${{ matrix.gemfile }}
       SCOUT_TEST_FEATURES: ${{ matrix.test_features }}
+      SCOUT_USE_PREPEND: ${{ matrix.prepend }}
 
     runs-on: ubuntu-latest
 
--- a/gems/instruments.gemfile
+++ b/gems/instruments.gemfile
@@ -1,0 +1,6 @@
+eval_gemfile("../Gemfile")
+
+gem 'httpclient'
+gem 'http'
+gem 'redis'
+gem 'moped'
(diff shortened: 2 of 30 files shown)


andychongyz:
Any timeline to get this merged? It seems to affect quite a lot of updated Rails app.

samsonjs:
We’ve been holding off on updating Scout for a year because of this issue. It’s also blocking us from updating to Ruby 3.1 because the old Scout gem isn’t compatible with the new YAML parser that’s safe by default, which disables references.

natematykiewicz:
Is there any update on this?

I have Scout + Sentry in my Gemfile. I installed Flipper and it caused stack level too deep errors on all Net::HTTP requests.

From what I can tell, the entire problem is that Flipper requires net/http.
https://github.com/jnunemaker/flipper/blob/f4a50467a6da84f4d886b8bec1df56127446002b/lib/flipper/adapters/http.rb#L1

I had a similar problem with Logtail a few months ago. That one requires net/https.
https://github.com/logtail/logtail-ruby/blob/63a7be22d22b0de4d373bf1967eeb4e562c67195/lib/logtail/log_devices/http.rb#L3

So when a gem requires net/http(s) before scout, bad things happen.

Prepending seems to be the solution here.

**This will Stack Level Too Deep during a Net::HTTP call**
```ruby
gem 'flipper'
gem 'flipper-active_record'
gem 'flipper-ui'

gem 'scout_apm'
gem 'sentry-ruby'
gem 'sentry-rails'
gem 'sentry-sidekiq'
```

**This works fine**
```ruby
gem 'scout_apm'
gem 'sentry-ruby'
gem 'sentry-rails'
gem 'sentry-sidekiq'

gem 'flipper'
gem 'flipper-active_record'
gem 'flipper-ui'
```

samsonjs:
Is Ruby not a priority for Scout anymore? It's amazing that this has been ready for review for 2 months and nothing has happened since then. I appreciate your work on this @dlanderson but I have to say that the timeline of this issue has me questioning Scout's commitment to Ruby.

AnilRh:
@dlanderson Do you have an ETA for this release? It sounded like you were hopeful to get this out late-June. Like others here our company uses both Scout and Airbrake - we are unable to upgrade due to the incompatibility between the two, so unless this is resolved soon we will be forced to pick a single service to move forward with. That's not an appealing option for us. 

Anything you can do to move this forward in a timely fashion is appreciated - thanks!

natematykiewicz:
Yeah, this is a pretty brutal bug. It's pretty disheartening to see that the fix has been sitting here for 3 months. It really does feel like Ruby isn't important to Scout.

dlanderson:
Hi all - Apologies for the delay. I had wanted to get more of the instrumentation set for prepend option with updated tests and clearly I haven't been able to get to it as soon as I wanted. I do have most of the instruments ready with prepend available, and I am planning to release a new major version next week.

samsonjs:
Thank you @dlanderson! Have a great weekend.

Issue #253, opened 2019-03-07
Scout causes Puma to run the on_worker_boot block twice


adriano-tirloni:
scout_apm (2.4.21)
Scout is causing Puma to execute the on_worker_boot twice, only when **preloading** is turned **on** on puma. This file (https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/server_integrations/puma.rb) gave me the hint by logging `I, [2019-03-07T16:47:33.844521 #186]  INFO -- : Installing Puma worker loop.` when I tought it was a Puma configuration issue. 
Production or Development env is the same.
If i remove the gem and bundle it stops running twice.

**preload off**
> [225] Puma starting in cluster mode...
> [225] * Version 3.12.0 (ruby 2.5.1-p57), codename: Llamas in Pajamas
> [225] * Min threads: 0, max threads: 1
> [225] * Environment: development
> [225] * Process workers: 1
> [225] * Phased restart available
> [225] * Listening on tcp://0.0.0.0:3000
> [225] Use Ctrl-C to stop
> --before_fork-start
> --before_fork-end
> worker 0=>--on_worker_fork--
> worker 0=>--after_worker_fork--
> worker 0=>--on_worker_boot--
> [225] - Worker 0 (pid: 228) booted, phase: 0
>...
> [225] - Gracefully shutting down workers...
> worker 0=>--on_worker_shutdown--
> [225] === puma shutdown: 2019-03-07 16:57:53 +0000 ===
> [225] - Goodbye!

**preload on**

> [275] Puma starting in cluster mode...
> [275] * Version 3.12.0 (ruby 2.5.1-p57), codename: Llamas in Pajamas
> [275] * Min threads: 0, max threads: 1
> [275] * Environment: development
> [275] * Process workers: 1
> [275] * Preloading application
> [275] * Listening on tcp://0.0.0.0:3000
> [275] Use Ctrl-C to stop
> --before_fork-start
> --before_fork-end
> worker 0=>--on_worker_fork--
> worker 0=>--after_worker_fork--
> **worker 0=>--on_worker_boot--
> I, [2019-03-07T17:01:58.601184 #278]  INFO -- : Installing Puma worker loop.
> worker 0=>--on_worker_boot--**
> [275] - Worker 0 (pid: 278) booted, phase: 0
> ...
> [275] - Gracefully shutting down workers...
> worker 0=>--on_worker_shutdown--
[...]

Issue #325, opened 2020-03-23
Avoid starting threads in the parent process of forking servers


cschneid:
A customer reports that Puma is emitting a warning during start that the background thread is being started and won't survive a fork.

This does not affect the behavior of Scout, as the thread will be restarted in the forked child process, but it is a scary looking warning that we should avoid.

```
12:21:16 PM web.1 |  [16321] ! WARNING: Detected 2 Thread(s) started in app boot:
12:21:16 PM web.1 |  [16321] ! #<Thread:0x00007fae721b32d0@/Users/.../.rbenv/versions/2.4.9/lib/ruby/gems/2.4.0/gems/scout_apm-2.6.6/lib/scout_apm/agent.rb:174 sleep> - /Users/.../.rbenv/versions/2.4.9/lib/ruby/gems/2.4.0/gems/scout_apm-2.6.6/lib/scout_apm/background_worker.rb:51:in `sleep'
```

dlanderson:
I can't reproduce with the latest `scout_apm_ruby` version 5.0.0. Please reopen if someone hits this again.

bf4:
fwiw, been seeing this for a while, currently on Ruby 2.7.6, scout_apm-5.1.1, 6.1.4.4

running on Heroku

```
2022-04-20T03:52:01.339485+00:00 app[web.2]: [3] * Listening on http://0.0.0.0:23787
2022-04-20T03:52:01.339526+00:00 app[web.2]: [3] ! WARNING: Detected 2 Thread(s) started in app boot:
2022-04-20T03:52:01.339564+00:00 app[web.2]: [3] ! #<Thread:0x00007ff9c2165148 /app/vendor/bundle/ruby/2.7.0/gems/scout_apm-5.1.1/lib/scout_apm/agent.rb:176 sleep> - /app/vendor/bundle/ruby/2.7.0/gems/scout_apm-5.1.1/lib/scout_apm/background_worker.rb:51:in `sleep'
2022-04-20T03:52:01.339587+00:00 app[web.2]: [3] ! #<Thread:0x00007ff9bfac0ee8 /app/vendor/bundle/ruby/2.7.0/gems/scout_apm-5.1.1/lib/scout_apm/agent.rb:210 sleep> - /app/vendor/bundle/ruby/2.7.0/gems/scout_apm-5.1.1/lib/scout_apm/background_worker.rb:51:in `sleep'
```

PR #463, opened 2022-11-22
Work around Puma config option issue causing duplicate before_worker_boot invocations when scout_apm is loaded

rubiety:
## Introduction

This fixes issue #253, which describes a bug causing Puma `on_worker_boot` callbacks to fire *twice* per definition when the `scout_apm` gem is included in the same Rails application. We discovered this issue (and then stumbled upon issue #253) while trying to add some statistics logging in our own Puma plugin, via a tread spawned in `on_worker_boot`. We were able to diagnose this as being an issue specifically with the `on_worker_boot` *callbacks* being fired twice (they were not *defined* or loaded twice), and it occurs only when `scout_apm` is included in the application and active via `monitor: true` in `config/scout_apm.yml`.

## Replication

This can be replicated by including both `puma` and `scout_apm` in a Rails application, setting `monitor: true` for your environment in `config/scout_apm.yml`, and then within `config/puma.rb` configuring two workers (since `on_worker_boot` is only invoked in multiple worker mode) as well as an `on_worker_boot` callback that logs a string:

```ruby
workers 2

on_worker_boot do
  puts "Invoked on_worker_boot"
end
```

This will result in the string "Invoked on_worker_boot" being logged *twice per worker*. If you remove `gem "scout_apm"` from the Gemfile, it will correctly be logged once per worker.

## The Issue

The underlying issue is a fairly nuanced interaction issue between how `scout_apm` hooks into Puma via `ScoutApm::ServerIntegrations::Puma#install` and how Puma handles getting and setting configuration options (which `scout_apm` uses via `::Puma.cli_config.options`).

Although the Scout integration appears to be straightforwardly grabbing an existing `before_worker_boot` value from `options`, appending to it, and re-setting the same key, Puma's implementation of configuration options (in `Puma::UserFileDefaultOptions`) unfortunately does not symmetrically handle `[]` and `[]=`:

```ruby
def [](key)
  fetch(key)
end

def []=(key, value)
  user_options[key] = value
end

def fetch(key, default_value = nil)
  return user_options[key]    if user_options.key?(key)
  return file_options[key]    if file_options.key?(key)
  return default_options[key] if default_options.key?(key)

  default_value
end
```

Note how setting will always set `user_options`, but getting is consulting three different underlying hashes in sequence. So if there is a `before_worker_boot` on `file_options` or `default_options`, it will be returned when consulting `options[:before_worker_boot]`. Scout then ppends to this array and sets it, intending to replace the existing array. But because `[]=` only sets `user_options` internally in `Puma::UserFileDefaultOptions`, we end up effectively with duplicated `before_worker_boot` handlers because of how scout interprets options in `run_hooks` and `all_of`:

```ruby
def run_hooks(key, arg, log_writer, hook_data = nil)
  @options.all_of(key).each do |b|
    ...
end

def all_of(key)
  user    = user_options[key]
  file    = file_options[key]
  default = default_options[key]

  user    = [user]    unless user.is_a?(Array)
  file    = [file]    unless file.is_a?(Array)
  default = [default] unless default.is_a?(Array)

  user.compact!
  file.compact!
  default.compact!

  user + file + default
end
```

## The Fix

This PR works around this issue by directly looking up `user_options`, which will ensure parity between getting and setting after appending. Because this is highly non-obvious unless you consult the Puma implementation of `options`, I've added an explanatory comment to the `install` method.

## Puma Plugin Implementation

If it's at all possible to do, it's probably more ideal for Scout to hook into Puma via a [puma plugin](https://github.com/puma/puma/blob/master/docs/plugins.md), which will avoid some of the awkwardness of manually fetching and re-setting existing options (that in part gave rise to this issue). That's out of scope for this PR, which is intent on working around the underlying issue described in original issue #253.

### DIFF
--- a/lib/scout_apm/server_integrations/puma.rb
+++ b/lib/scout_apm/server_integrations/puma.rb
@@ -23,8 +23,36 @@ module ScoutApm
         defined?(::Puma) && (File.basename($0) =~ /\Apuma/)
       end
 
+      # Puma::UserFileDefaultOptions exposes `options` based on three underlying
+      # hashes: user_options, file_options, and default_options. While getting an `options`
+      # key consults all three underlying hashes, setting an `options` key only sets the
+      # user_options hash:
+      #
+      #    def [](key)
+      #      fetch(key)
+      #    end
+      #
+      #    def []=(key, value)
+      #      user_options[key] = value
+      #    end
+      #
+      #    def fetch(key, default_value = nil)
+      #      return user_options[key]    if user_options.key?(key)
+      #      return file_options[key]    if file_options.key?(key)
+      #      return default_options[key] if default_options.key?(key)
+      #
+      #      default_value
+      #    end
+      #
+      # Because of this, we can't read options[:before_worker_boot], modify, and then re-set
+      # options[:before_worker_boot], since doing so could cause duplication if `before_worker_boot`
+      # exists on the other two underlying hashes (file_options, default_options).
+      #
+      # To get around this, we explicitly read from `user_options` only, and still set using `options[]=`,
+      # which Puma allows for setting `user_options`.
+      #
       def install
-        old = ::Puma.cli_config.options[:before_worker_boot] || []
+        old = ::Puma.cli_config.options.user_options[:before_worker_boot] || []
         new = Array(old) + [Proc.new do
           logger.info "Installing Puma worker loop."
           ScoutApm::Agent.instance.start_background_worker


dlanderson (approved):
Thanks so much for the detailed comments and PR! 👍

summera:
@dlanderson Please hold off on a merge for just a bit. We're running into an issue with these changes when testing. I'll leave a comment once we've resolved.

summera:
@dlanderson should be all set now. We were attempting to access `user_options` on a `Puma::Configuration` instance rather than on an instance of `Puma::UserFileDefaultOptions`. This seems to be working for us, but please let us know if you see any issues. As a side note, it would be great if there were some tests added for these server integrations (like Puma) as a follow-up to ensure compatibility and avoid regressions going forward. It's possible that this could break without us knowing if Puma were to release a backwards incompatible change related to `Puma::Configuration`. Thanks!

bf4:
Closes https://github.com/scoutapp/scout_apm_ruby/issues/325

PR #465, opened 2022-12-15
Fix: call `add_instruments` after rails initialization when rails version is >= 3

gustavothecoder:
I have an initializer with this config in my Rails 6 app:
```ruby
# Some configurations...

Rails.application.config.active_record.belongs_to_required_by_default = false

# More configurations...
```
After adding the `scout_apm_ruby` to my project, this config started not loading, which broke some features in my app. After investigating, I found that `scout_apm_ruby` was starting `ActiveRecord` without my config, this is happening because `install_via_after_initialize?` never returns true when Rails version is different from 3, which is a bug since all Rails versions above 3 can respond to `configuration` as well.

In this PR I fix this bug and add some tests to avoid regression.

### DIFF
--- a/lib/scout_apm/instruments/active_record.rb
+++ b/lib/scout_apm/instruments/active_record.rb
@@ -66,7 +66,7 @@ module ScoutApm
         defined?(::Rails) &&
           defined?(::Rails::VERSION) &&
           defined?(::Rails::VERSION::MAJOR) &&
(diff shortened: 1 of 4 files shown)


jrothrock:
Thanks for the pull request!

jrothrock (approved):

PR #466, opened 2023-01-27
Adds Ruby 3.1 and 3.2. Quotes 3.0 to avoid truncation. Updates action versions

petergoldstein:
Quoting '3.0' is necessary to avoid truncation to '3'.  If truncated to '3', then the latest Ruby 3 (currently Ruby 3.2.0) is loaded, which is not the intended behavior.


mjobin-mdsol (approved):

jrothrock (approved):
Thank you for the PR!

PR #468, opened 2023-03-15
Add back PercentPolicy to SlowRequestPolicy

ibrahima:
I was noticing that I'm not getting a lot of traces for high-traffic endpoints, and I think this might be why? I think this PercentPolicy got lost in a refactor, and PercentilePolicy was listed twice instead. It does look like [PercentPolicy](https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/slow_policy/percent_policy.rb) is the rule that helps track high traffic endpoints.

ref: https://github.com/scoutapp/scout_apm_ruby/commit/026f04ea77dde448da9d82b7456abbd849f5f315

### DIFF
--- a/lib/scout_apm/slow_request_policy.rb
+++ b/lib/scout_apm/slow_request_policy.rb
@@ -14,7 +14,7 @@ module ScoutApm
 
     def add_default_policies
       add(SlowPolicy::SpeedPolicy.new(context))
-      add(SlowPolicy::PercentilePolicy.new(context))
+      add(SlowPolicy::PercentPolicy.new(context))
       add(SlowPolicy::AgePolicy.new(context))
       add(SlowPolicy::PercentilePolicy.new(context))
     end


mjobin-mdsol (approved):

jrothrock (approved):
Great catch! Thank you for the PR!

ibrahima:
Haha thanks! I had noticed that most of my traces were for slow endpoints rather than high traffic endpoints so I dug into the gem to see how it decides to sample a trace.

I did actually put my forked version into production and I don't seem to see a difference (or at least I am still missing traces for my highest throughput endpoints and rather see a lot of the slower ones instead) so I'm not sure if there's some reasons why that wouldn't work or if there's more at play here... I guess the relative scales of the various score factors might also be a factor?

jrothrock:
Yeah it'll be based on the relative factors. If an endpoint happens to not have reported for awhile, and has a slow trace, the combination of the age and speed can be really quite high.

We do have the ability to add a custom request policies, if you're looking to say duplicate the PercentPolicy and give it a bit more weight:
https://scoutapm.com/docs/ruby/in-depth#request-policies

ibrahima:
Thanks! That's really helpful, I'll try that and see if that helps. For now I'm just trying the suggestion there of adding weight to specific endpoints, though duplicating PercentPolicy with a higher weight does seem more sustainable long term.

This kinda leaves me a little more confused than before though. Here's what I understood of the current policies:

- AgePolicy: `(minutes_since_last_traced) * 0.25`
- PercentPolicy: `percent_of_total_time * 2.5`
- PercentilePolicy: piecewise function that returns
  - 0.4 if p < 40 (pretty fast)
  - 1.4 if 40 < p < 60 (average)
  - 0.7 if 60 < p < 90 (above average but not exceptionally so)
  - 1.4 if p > 90 (really slow)
- SpeedPolicy: `Math.log(1 + total_time_seconds) * 0.25`

With those policies, it seems like most of these will cap out at some value less than 2 for realistic numbers. PercentPolicy would probably be ~0.25 for a high-traffic endpoint (our top endpoints are around ~10% total time in our case). PercentilePolicy has fixed values so it would top out at 1.4 for either an average request or a really slow request. SpeedPolicy grows pretty slowly, it takes a 50s request to get even close to 1: `Math.log(1 + 50) * 0.25 => 0.9829564081810814`. That leaves AgePolicy, which has no caps or logarithmic growth, so in theory an endpoint that we haven't seen in 8 minutes would get a score of 2 from AgePolicy, and that term would quickly dominate the other terms. But if that's the case, I would expect the final behavior to roughly round-robin sample the endpoints in the long run, because that term would dominate. But the behavior I'm actually seeing is that our high traffic endpoints (~10% of total time) will very often get no traces for a whole day or more, whereas some other endpoints get multiple traces in the last 30 minutes. It's really perplexing, because before doing the math I would have thought that it was prioritizing slow requests above all else, but I see some endpoints that are low traffic (rounds to 0.0% total time) and also quite fast (under 20ms) that have multiple samples in every period, whereas the endpoints that are higher traffic have none... huh. 🤔

I know that we do restart our workers via unicorn-worker-killer if they use too much memory, but I think the restarts should not happen so frequently that the AgePolicy counters are getting reset to the point of not having an effect. I just really don't understand why I'm getting so many traces from an endpoint that's so fast and rare and zero traces from a common, average endpoint 😕.

From looking at the logs, it seems like on each server, the Scout APM agent will record 10 traces per minute. I'm not sure if there's some higher level aggregation that happens after that point that narrows it down further?

Hopefully once we deploy the custom policy we'll be able to see what magnitude of score bump affects what endpoints get sampled. Perhaps Scout's support team can help debug this further looking at our production data.

Anyway, thanks for your help, this has been very enlightening!

PR #469, opened 2023-03-20
Add test targets and delete EOL Ruby version test target

kentarohorie:


.github/workflows/test.yml
```diff
@@ -19,21 +19,6 @@ jobs:
...
-          - ruby: 2.4
-          - ruby: 2.5
-          - ruby: 2.6
-          - ruby: 2.6
-            gemfile: gems/typhoeus.gemfile
-            test_features: "typhoeus"
-          - ruby: 2.6
-            gemfile: gems/octoshark.gemfile
-          - ruby: 2.6
-            gemfile: gems/rails3.gemfile
```
mjobin-mdsol, line 35: shall we also delete those old gemfiles ?

and bump the minimum requirement in the `scout_apm_ruby.gemspec` ?

.github/workflows/test.yml
```diff
@@ -50,6 +35,32 @@ jobs:
           - ruby: 3.0
```
mjobin-mdsol, line 35: 3.0 needs to be quoted `"3.0"` otherwise, latest 3.x is used

lancetarn:
Thanks for the concept and effort here. We will drop these tests and compatibility when we make a major version release.

mitchh456:
We did a smaller version of this dropping Rubies <= 2.4, and up to 3.3 to the matrix.

PR #486, opened 2024-01-19
Fix SyntaxError when instrumenting hash with shorthand

### LINKED ISSUE #445: Parser gem required in Gemfile
I was wondering why auto instruments wasn't working and it was because of the changes made in this PR.
https://github.com/scoutapp/scout_apm_ruby/pull/440

This PR checks whether `Parser::TreeRewriter` is defined and bails out early if it isn't.

The problem is for applications that do not explicitly include `parser` in there Gemfile it will not yet be defined and will never reach the [codepath where Scout requires it](https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/auto_instrument/parser.rb#L2)

The workaround is to explicitly add `parser` to the application Gemfile. Recommend updating docs to include this or reverting/changing the above PR.

### LINKED ISSUE #484: Auto-Instrumentation creates syntax error
The auto-instrumentation patches do not work with Ruby 3 hashes that do not have a value.

This is perfectly valid Ruby 3.2.2 code in a Rails controller:
```ruby
def base_cache_key
  {
    path: request.env['PATH_INFO'],
    timestamp: (Time.now.to_i / 300),
    publishing_status:,
  }
end
```

Which causes this syntax error when deployed to production:

```
Nov 3 03:00:19 PM  bundler: failed to load command: puma (/app/vendor/bundle/ruby/3.2.0/bin/puma)
Nov 3 03:00:19 PM  /app/vendor/bundle/ruby/3.2.0/gems/scout_apm-5.3.5/lib/scout_apm/auto_instrument/instruction_sequence.rb:11:in `compile': /app/app/controllers/api/v4/experiences_controller.rb:86: syntax error, unexpected ':', expecting => (SyntaxError)
Nov 3 03:00:19 PM  ...che_key'"]){publishing_status}:,
Nov 3 03:00:19 PM  ...                              ^
Nov 3 03:00:19 PM
Nov 3 03:00:19 PM       from /app/vendor/bundle/ruby/3.2.0/gems/scout_apm-5.3.5/lib/scout_apm/auto_instrument/instruction_sequence.rb:11:in `load_iseq'
```

The workaround is to add a value to the Hash, even though it's not necessary anymore.

```ruby
def base_cache_key
  {
    path: request.env['PATH_INFO'],
    timestamp: (Time.now.to_i / 300),
    publishing_status: publishing_status,
  }
end
```

kxmbrian:
# Why

We get a `SyntaxError` when instrumenting hashes with [Ruby 3.1 hash shorthand syntax](https://rubyreferences.github.io/rubychanges/3.1.html#values-in-hash-literals-and-keyword-arguments-can-be-omitted) in controllers.

Fixes #484
Fixes #445

# Fix

Add a `on_hash` method that does the instrumentation correctly. 

@jrothrock - would you be the right person to ask for a review? Do let me know if there's a cleaner way to do this.

# Testing

I've added test cases.

@natematykiewicz - if you would like to, you may test this with your app.

### DIFF
--- a/scout_apm.gemspec
+++ b/scout_apm.gemspec
@@ -28,10 +28,10 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rake-compiler"
   s.add_development_dependency "addressable"
   s.add_development_dependency "activesupport"
-  s.add_runtime_dependency "parser"
+  s.add_runtime_dependency "parser", Gem::Version.new(RUBY_VERSION).approximate_recommendation
 
   # These are general development dependencies which are used in instrumentation
-  # tests. Specific versions are pulled in using specific gemfiles, e.g. 
+  # tests. Specific versions are pulled in using specific gemfiles, e.g.
   # `gems/rails3.gemfile`.
   s.add_development_dependency "activerecord"
   s.add_development_dependency "sqlite3"
--- a/lib/scout_apm/auto_instrument/rails.rb
+++ b/lib/scout_apm/auto_instrument/rails.rb
@@ -135,6 +135,27 @@ module ScoutApm
           wrap(node.location.expression, *instrument(node.location.expression.source, file_name, line))
         end
 
+        def on_hash(node)
+          node.children.each do |pair|
+            # Skip `pair` if we're sure it's not using the hash shorthand syntax
+            next if pair.type != :pair
+            key_node, value_node = pair.children
+            next unless key_node.type == :sym && value_node.type == :send
+            key = key_node.children[0]
+            next unless value_node.children[0].nil? && key == value_node.children[1]
+
+            # Extract useful metadata for instrumentation:
+            line = pair.location.line || 'line?'
+            # column = pair.location.column || 'column?' # not used
+            # method_name = key || '*unknown*' # not used
+            file_name = @source_rewriter.source_buffer.name
+
+            instrument_before, instrument_after = instrument(pair.location.expression.source, file_name, line)
+            replace(pair.loc.expression, "#{key}: #{instrument_before}#{key}#{instrument_after}")
+          end
+          super
+        end
+
         # def on_class(node)
         #   class_name = node.children[1]
         #
--- a/test/unit/auto_instrument/hash_shorthand_controller-instrumented.rb
+++ b/test/unit/auto_instrument/hash_shorthand_controller-instrumented.rb
@@ -1,0 +1,36 @@
+
+class HashShorthandController < ApplicationController
+  def hash
+    json = {
+      static: "static",
+      shorthand: ::ScoutApm::AutoInstrument("shorthand:",["ROOT/test/unit/auto_instrument/hash_shorthand_controller.rb:6:in `hash'"]){shorthand},
+      longhand: ::ScoutApm::AutoInstrument("longhand: longhand",["ROOT/test/unit/auto_instrument/hash_shorthand_controller.rb:7:in `hash'"]){longhand},
+      longhand_different_key: ::ScoutApm::AutoInstrument("longhand",["ROOT/test/unit/auto_instrument/hash_shorthand_controller.rb:8:in `hash'"]){longhand},
+      hash_rocket: ::ScoutApm::AutoInstrument(":hash_rocket => hash_rocket",["ROOT/test/unit/auto_instrument/hash_shorthand_controller.rb:9:in `hash'"]){hash_rocket},
+      :hash_rocket_different_key => ::ScoutApm::AutoInstrument("hash_rocket",["ROOT/test/unit/auto_instrument/hash_shorthand_controller.rb:10:in `hash'"]){hash_rocket},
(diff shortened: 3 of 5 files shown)


jrothrock:
Thanks for the PR and the included test cases!

I stepped through this with a locally reproduced issue of this, and it fixed it and looks good!

It does look like this requires `parser` version `3.0.3.0` and above, maybe we need to mention this in the docs, but with the general parser warning logs related to the Ruby version maybe we're good?

jrothrock (approved):

jrothrock:
I'm not sure this is something we can do right now unfortunately. 

We still support Ruby versions 2.1 - 2.7, even though they have been EOL, and `parser` [may not maintain backward compatibility for older Ruby versions](https://github.com/whitequark/parser?tab=readme-ov-file#backwards-compatibility) and throws a warning about version compliance

kxmbrian:
[Since](https://github.com/whitequark/parser?tab=readme-ov-file#backwards-compatibility):

> Parser versions are structured as `x.y.z.t`, where `x.y.z` indicates the most recent supported Ruby release (support for every Ruby release that is chronologically earlier is implied), and t is a monotonically increasing number.

would it be reasonable to do this?

```ruby
s.add_runtime_dependency "parser", ">= #{RUBY_VERSION}.0"
```

kxmbrian:
Pushed a commit with this instead because not every patch-level Ruby version has a corresponding `parser` gem version:

```rb
s.add_runtime_dependency "parser", Gem::Version.new(RUBY_VERSION).approximate_recommendation
```

Looks better without any string manipulation and the tests pass.

natematykiewicz:
I'll bet this explains why I've had auto instrumentation not work at times and then later magically work -- I had upgraded Ruby without upgrading Parser, resulting in Parser being incompatible with my Ruby version. Eventually I update some gem that relies on Parser (Scout, Rubocop) and get a version of Parser that actually supports my Ruby version and auto instrumentation starts working again.

scout_apm.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rake-compiler"
   s.add_development_dependency "addressable"
   s.add_development_dependency "activesupport"
-  s.add_runtime_dependency "parser"
+  s.add_runtime_dependency "parser", Gem::Version.new(RUBY_VERSION).approximate_recommendation
```
natematykiewicz, line 31: Per https://bundler.io/guides/gemfile.html

> Most of the version specifiers, like >= 1.0, are self-explanatory. The specifier ~> has a special meaning, best shown by example. ~> 2.0.3 is identical to >= 2.0.3 and < 2.1. ~> 2.1 is identical to >= 2.1 and < 3.0. ~> 2.2.beta will match prerelease versions like 2.2.beta.12. ~> 0 is identical to >= 0.0 and < 1.0.

```suggestion
  s.add_runtime_dependency "parser", "#{Gem::Version.new(RUBY_VERSION).approximate_recommendation}.0"
```

Since Parser has a version that corresponds to every minor version, we want `~> 3.2.0` which means `>= 3.2.0 and < 3.3`. Right now you've got `~> 3.2` which means `>= 3.2 and < 4.0`. But I don't think we want to use Parser 3.3 on Ruby 3.2.

natematykiewicz (changes requested):

scout_apm.gemspec
kxmbrian: Good catch, thanks!

kxmbrian: Hmm getting these errors for the tests

```
warning: parser/current is loading parser/ruby23, which recognizes
warning: 2.3.3-compliant syntax, but you are running 2.3.8.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
rake aborted!
LoadError: Parser::TreeRewriter was not defined
```

```
warning: parser/current is loading parser/ruby24, which recognizes
warning: 2.4.0-compliant syntax, but you are running 2.4.10.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
rake aborted!
LoadError: Parser::TreeRewriter was not defined
```

[The last Parser 2.3 is Parser 2.3.3.1](https://github.com/whitequark/parser/blob/master/CHANGELOG.md). Looks like we're expected to use Parser 2.4 for Ruby 2.3.8.

> But I don't think we want to use Parser 3.3 on Ruby 3.2.

Agree with you that this is not ideal, but it's probably safe enough. The main thing is to get folks to upgrade the parser whenever they upgrade their Ruby version.

Gonna revert for now.

jrothrock: Thanks for the continued effort on this guys. We appreciate it!

I think this is good.

This also may solve some issues related to:
https://github.com/scoutapp/scout_apm_ruby/issues/445

But still allows people below Ruby 2.5 to get to a parser version above 2.5 which has the TreeRewriter. Again also not ideal, but probably safe

jrothrock (approved):

PR #488, opened 2024-01-30
Fix nested_call edge case

kxmbrian:
# What

@jrothrock @natematykiewicz - I missed an edge case in #486

The `replace` action in `on_hash` should ideally run only for the case where the Ruby 3.1 hash shorthand syntax is used. If the `value_node` takes in an argument, we can be sure it's not the shorthand syntax case. This PR adds a check for that: 

```rb
value_node.children.size == 2
```

# Test

Without the change, `params["timestamp"]` gets dropped:

```rb
# Prior to this PR
nested_call: ::ScoutApm::AutoInstrument(\"nested_call: nested_call(params[\\\"timestamp\\\"])\",[\"ROOT/test/unit/auto_instrument/hash_shorthand_controller.rb:15:in `hash'\"]){nested_call}
# Fixed
nested_call: ::ScoutApm::AutoInstrument(\"nested_call(params[\\\"timestamp\\\"])\",[\"ROOT/test/unit/auto_instrument/hash_shorthand_controller.rb:15:in `hash'\"]){nested_call(params[\"timestamp\"])}
```

leading to errors like `wrong number of arguments (given 0, expected 1) (ArgumentError)`.

### DIFF
--- a/lib/scout_apm/auto_instrument/rails.rb
+++ b/lib/scout_apm/auto_instrument/rails.rb
@@ -142,7 +142,7 @@ module ScoutApm
             key_node, value_node = pair.children
             next unless key_node.type == :sym && value_node.type == :send
             key = key_node.children[0]
-            next unless value_node.children[0].nil? && key == value_node.children[1]
+            next unless value_node.children.size == 2 && value_node.children[0].nil? && key == value_node.children[1]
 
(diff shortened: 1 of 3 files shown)


jrothrock:
Nice catch!

jrothrock (approved):

PR #494, opened 2024-03-01
Don't capture queue time for websocket requests

jrothrock:
With the use of libraries such as Stimulus Reflex, which under the hood utilize Turbo Streams, we can run into issues where the queue times that we report are very inaccurate.

They appear to have the same queue time header/timestamp as when the connection was first created, and is not changed for subsequent data sent over the websocket. This leads to an ever growing queue time as long as the connection isn't terminated.


lancetarn (approved):

mitchh456 (approved):

Zajn:
Hey y'all, thanks for fixing and merging this ❤️ Our scout graphs look so much better now.

Any chance there will be a minor version bump to the gem that includes this?

jrothrock:
Hey @Zajn, sorry for the delay on this. We just released `5.3.8` with this change

Zajn:
@jrothrock No worries, thanks!

Issue #495, opened 2024-03-22
Frozen string literals?


ibrahima:
Hi! I've been having some memory issues with an application and I started profiling it using https://github.com/MiniProfiler/rack-mini-profiler#memory-profiling. I found the source of the leaks, but I also noticed that in the allocated strings section the Scout APM gem comes up a lot (at the top of the list). It _seems_ like we could save on a lot of memory allocations by freezing some string literals and/or creating constants for commonly used strings.

```

      2868  "AutoInstrument"
      1459  /rails/vendor/ruby/3.2.0/gems/scout_apm-5.3.5/lib/scout_apm/tracked_request.rb:192
      1391  /rails/vendor/ruby/3.2.0/gems/scout_apm-5.3.5/lib/scout_apm/auto_instrument/layer.rb:9
        12  /rails/vendor/ruby/3.2.0/gems/scout_apm-5.3.5/lib/scout_apm/bucket_name_splitter.rb:17
         6  /rails/vendor/ruby/3.2.0/gems/scout_apm-5.3.5/lib/scout_apm/instruments/action_controller_rails_3_rails4.rb:102

      1394  ":"
      1391  /rails/vendor/ruby/3.2.0/gems/scout_apm-5.3.5/lib/scout_apm/auto_instrument/layer.rb:6
         3  /usr/local/lib/ruby/3.2.0/net/http.rb:1986
```
https://github.com/scoutapp/scout_apm_ruby/blob/bf9fb4b55fc2e2412a76a4d8731b55bb3b1d7e15/lib/scout_apm/tracked_request.rb#L192
https://github.com/scoutapp/scout_apm_ruby/blob/bf9fb4b55fc2e2412a76a4d8731b55bb3b1d7e15/lib/scout_apm/auto_instrument/layer.rb#L9

https://github.com/scoutapp/scout_apm_ruby/blob/bf9fb4b55fc2e2412a76a4d8731b55bb3b1d7e15/lib/scout_apm/auto_instrument/layer.rb#L6 also comes up a lot - I guess the file names get computed a lot and the colon is also reallocated frequently.

I'm not sure if it's enough to stick a `.freeze` after the literals or if you need to pull out a constant to save the memory allocations.

Maybe for a bigger application this would just be noise, but it happens that the "AutoInstrument" string is the top string allocated in the memory profiles that I'm looking at. I think overall Scout's overhead is pretty low but it would always be good to get it lower!

PR #496, opened 2024-03-22
Freeze string literals in AutoInstrument layer

ibrahima:
I think if you have a lot of AutoInstrumentation going on, this results in a lot of repeated memory allocations for the same string.

Addresses #495

I tested this out in our staging environment and it does remove the string allocations from strings in these files. I don't know if this is meaningful at all or if it's worth doing it piecemeal like this, versus just adding the pragma to everything or relying on the user to pass some Ruby options to freeze all string literals.


ibrahima:
Replaced by #536

PR #498, opened 2024-05-06
Fix deprecation warning in ActiveRecord 7.1 when clearing active connections

haffla:
Hi. Passing no argument to `ActiveRecord::Base.connection_handler.clear_active_connections!` was deprecated [here](https://github.com/rails/rails/pull/45924) and produces following log.

```
DEPRECATION WARNING: `clear_active_connections!` currently only applies to connection pools in the current role (`writing`). In Rails 7.2, this method will apply to all known pools, regardless of role. To affect only those connections belonging to a specific role, pass the role name as an argument. To switch to the new behavior, pass `:all` as the role name.
```

Rails documentation of method: https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/ConnectionHandler.html#method-i-clear_active_connections-21

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
@@ -23,7 +23,9 @@ jobs:
             gemfile: gems/rails3.gemfile
           - ruby: 2.2
           - ruby: 2.3
+            gemfile: gems/sqlite3-1.3.gemfile
           - ruby: 2.4
+            gemfile: gems/sqlite3-1.3.gemfile
           - ruby: 2.5
           - ruby: 2.6
           - ruby: 2.6
@@ -31,9 +33,6 @@ jobs:
             test_features: "typhoeus"
           - ruby: 2.6
             gemfile: gems/octoshark.gemfile
-          - ruby: 2.6
-            gemfile: gems/rails3.gemfile
-            bundler: 1.17.3
           - ruby: 2.7
           - ruby: 2.7
             prepend: true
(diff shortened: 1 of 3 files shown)


haffla:
Failing Github actions don't seem to be related to my change.

lancetarn:
Hello @haffla ! Thanks for putting this up. We'll have to see what changed out from under us to cause those test failures.

haffla:
Hi @lancetarn. Were you able to dig into this?

haffla:
The issue seems to be

```
LoadError: Error loading the 'sqlite3' Active Record adapter. Missing a gem it depends on? can't activate sqlite3 (~> 1.4), already activated sqlite3-2.0.1. Make sure all dependencies are added to Gemfile.
```

Rails seems to have had the [same issue](https://github.com/rails/rails/pull/51592).

haffla:
@lancetarn all checks passed now

lib/scout_apm/app_server_load.rb
```diff
@@ -51,7 +51,7 @@ def data
     ensure
       # Sometimes :database_engine and :database_adapter can cause a reference to an AR connection.
       # Make sure we release all AR connections held by this thread.
-      ActiveRecord::Base.connection_handler.clear_active_connections! if Utils::KlassHelper.defined?("ActiveRecord::Base")
+      ActiveRecord::Base.connection_handler.clear_active_connections!(:all) if Utils::KlassHelper.defined?("ActiveRecord::Base")
```
marianposaceanu, line 54: Hello - I wanted to open the same PR (when running tests this deprecation warning gets quite verbose). In any case I think this is a safer way to implement this (backwards compatibility):

```ruby
      if ActiveRecord::Base.connection_handler.respond_to?(:clear_active_connections!)
        ActiveRecord::Base.connection_handler.clear_active_connections!(:all)
      elsif ActiveRecord::Base.respond_to?(:clear_active_connections!)
        # ~> Rails 3.2 -> https://apidock.com/rails/ActiveRecord/Base/clear_active_connections%21/class
        ActiveRecord::Base.clear_active_connections!
      end
```

haffla: Ah you mean that `clear_active_connections!` doesn't take any arguments in Rails 3? That could be actually. But I wonder why no test is failing. I am gonna look into it.

haffla: Oh lol, this code isn't even tested?!

Issue #336, opened 2020-06-12
Allow ActiveJob before_perform callbacks to call ignore! and similar Scout APIs


cschneid:
A customer running Sidekiq via ActiveJob reports that calling ignore from a before_perform callback is not working. Calling the ignore from inside the job's perform method appears to be fine however.

The sequence a task gets executed appears to be:

* Sidekiq runner pops off a job, and calls 'process'
* Instantiates a ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper
* Runs server_middleware array
* Calls perform
* ActiveJob::Base.execute is called (defined in execution.rb).
* Execute callbacks are run
* The wrapped job is unwrapped, and perform_now is called (also defined in ActiveJob's execution.rb)
* Perform callbacks are run
* Job#perform is finally called

Scout injects itself as a server_middleware, so we should have the state necessary to ignore or rename or whatever API call you need by then.

dzirtusss:
Same here. This bug was misleading, as `before_perform` is the right place to put such a logic. At least keeping common configuration in `ApplicationJob` with `before_perform` looks much easier vs inserting it into all `perform` methods.

@cschneid thanks for reporting this, as it is good to know that this is a gem bug that smb else has, and not code bug.

knu:
Thanks for this piece of information.

Here's our workaround:

```ruby
# app/concerns/apm_sampling
module ApmSampling
  extend ActiveSupport::Concern

  private def apm_sample
    ScoutApm::Transaction.ignore! unless apm_sample?
  end

  private def apm_sample?
    true
  end

  def perform_with_apm_sampling(*args)
    apm_sample
    perform_without_apm_sampling(*args)
  end

  included do
    case
    when self < ActiveJob::Base
      def self.method_added(method_name)
        case method_name
        when :perform
          # Avoid reentrance
          unless method_defined?(:perform_without_apm_sampling)
            alias_method :perform_without_apm_sampling, :perform
            alias_method :perform, :perform_with_apm_sampling # This invokes method_added(:perform)
          end
        end
      end
    when self < ActionController::Base
      before_action :apm_sample
    when self < ActionController::Metal
      # There is no before_action callback in Metal.
      # You need to manually call apm_sampling from within each action method.
    end
  end
end

# app/controllers/application_controller.rb
class ApplicationController < ActionController::Base
  include ApmSampling

  private def apm_sample?
    rand < 0.05
  end
end

# app/jobs/application_job.rb
[...]

PR #501, opened 2024-07-05
Respect ignored requests on find

camallen:
ensure we can set a request to be ignored in a background processing system via a before hook and the request manager should return the ignored request instead of creating a new one

related issues https://github.com/scoutapp/scout_apm_ruby/issues/336

### DIFF
--- a/lib/scout_apm/request_manager.rb
+++ b/lib/scout_apm/request_manager.rb
@@ -12,7 +12,6 @@ module ScoutApm
       req = Thread.current[:scout_request]
 
       # this ordering is important as if the req is set to be ignored it's also stopping and recorded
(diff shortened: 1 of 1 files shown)


lancetarn:
Hello! This seems like the right thing to be doing! Thank you for the time and effort. We are working on managing the dependencies causing these test failures and will get this merged soon.

camallen:
This PR doesn't work as expected and is not currently mergeable.

With this change the `RequestManager` reuses the ignored tracked request and it never clears after a job is processed. I'm still looking at public APIs or a new reset API to enforce the ignored request clears correctly.

camallen:
Closing this PR for now to keep things tidy

PR #503, opened 2024-07-09
allow sqlite3 gem to float to version 2

neumayr:
testing…


lancetarn:
It seems like our backwards compatibility has finally become untenable. We will need to do some additional research to see what we can reasonably do. Thanks for the time on this.

PR #500, opened 2024-07-01
fix deprecation warning Bundler.rubygems.all_specs

neumayr:
- fix deprecation warning when starting `rails server`
  - `[DEPRECATED] Bundler.rubygems.all_specs has been removed in favor of Bundler.rubygems.installed_specs`
- reproducible with ruby v3.3.3+ (and new minimum required bundler v2.5.11, rubygems v3.5.11 versions)

#### Link
- https://github.com/rubygems/rubygems/pull/7673

Fixes #513 

/cc @mpvosseller

PR #516, opened 2024-12-10
fix failing test with ruby 3.2 and 3.3

neumayr:
Enhance CI configuration by adding sqlite3 gemfile for Ruby 3.2/3.3 in test matrix

🤷 Sorry, I have no idea why this change has an impact on the old failing ruby tests…

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,7 +44,9 @@ jobs:
             test_features: "sidekiq_install"
           - ruby: 3.1
           - ruby: 3.2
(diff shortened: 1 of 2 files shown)


mitchh456:
Hey @neumayr, thanks for opening this. I don't think your changes here are causing the other failures. I'm seeing the same in another PR. I think we're going to try to merge this and #500 soon, as well as fix those other new failing checks.

neumayr:
@mitchh456 thanks, rebase is done. 🚀

mitchh456 (approved):

PR #524, opened 2025-01-07
add sampling and ignore updates to changelog

mitchh456:


quinnmil (approved):
I figured we'll be updated documentation as well. The Python agent changelog links to documentation in a few places, so probably not a bad idea to do the same here once these are out of "unreleased"

PR #563, opened 2025-09-04
Add OpenSearch instrumentation support

dlackty:
- Add OpenSearch::Transport::Client instrumentation
- Support same operations as Elasticsearch integration
- Auto-detect OpenSearch gem presence
- Track operations under 'OpenSearch' category


lancetarn:
What a delight! Looks like it will do the trick, we will get this out shortly. Definitely 1000x better than a support ticket!

dlackty:
We have deployed this to production (app ID: 19), and it works well!

PR #567, opened 2025-09-18
564 float sample rate

### LINKED ISSUE #564: Allow sample_rate to support floats or callable values
Currently, sample_rate only supports integer values ([config implementation](https://github.com/scoutapp/scout_apm_ruby/blob/82d8d33312773b9983e0b5006af4b9129ef0a00d/lib/scout_apm/config.rb#L47-L51)).

For high-traffic applications or apps with many background workers, having only integer-based percentages makes sampling too coarse. For example, the minimum possible rate today is 1%, which is still too high in some environments.

**Proposal:**
* Allow sample_rate to accept floats (e.g., 0.001 = 0.001%).
  * Optionally, allow sample_rate to be set to a callable/closure, giving applications more flexibility (e.g., dynamically adjusting based on workload or environment).

**Example use case:**
We want to set the sample rate to 0.001% to avoid overwhelming Scout with data while still collecting representative samples.

**Benefits:**
* Finer-grained control over performance monitoring.
* Prevents excessive overhead in large-scale or background-heavy applications.
* Flexible configuration for diverse workloads.

How does this suggestion sound?

lancetarn:
Closes #564

### DIFF
--- a/lib/scout_apm/config.rb
+++ b/lib/scout_apm/config.rb
@@ -46,11 +46,11 @@ require 'scout_apm/environment'
 #                            instruments listed in this array. Default: []
 # ignore_endpoints         - An array of endpoints to ignore. These are matched as regular expressions. (supercedes 'ignore')
 # ignore_jobs              - An array of job names to ignore.
-# sample_rate              - Rate to sample entire application. An integer between 0 and 100. 0 means no traces are sent, 100 means all traces are sent.
-# sample_endpoints         - An array of endpoints to sample. These are matched as regular expressions with individual sample rate of 0 to 100.
-# sample_jobs              - An array of job names with individual sample rate of 0 to 100.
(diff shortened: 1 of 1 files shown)


mitchh456 (approved):

peterkleinct:
Hey, just want to say I wish the float clamping had been highlighted in the changelog. I saw support for floats mentioned but didn't think anything of it until our traces overran our daily quota today. I also spent a good few minutes wondering if we suddenly queued an extra 4 million jobs somehow. 😅 

I agree it's a good change, but should have been a lot more clear in the changelog that action was needed when upgrading.

lancetarn:
@peterkleinct That's true and I apologize. If this is causing ongoing problems for you, shoot us an email support @ scoutapm.com and we will find a way to make it right.

peterkleinct:
> @peterkleinct That's true and I apologize. If this is causing ongoing problems for you, shoot us an email support @ scoutapm.com and we will find a way to make it right.

Thanks - don't think it'll be a problem, I just updated our sample rates to 0.1 and 0.5 where appropriate. We're just setting them via heroku env so easy to change, just got a shock with the number of jobs sampled today :)

PR #583, opened 2025-11-11
Add NullableSampleRateCoercion for job_sample_rate

kimegede:
User who have already defined `job_sample_rate` before would end up with inconsistent behaviors.


lancetarn (dismissed):

lancetarn:
Thanks for catching this.

PR #610, opened 2026-02-13
Rage instrumentation

lancetarn:


lancetarn:
@rsamoilov thank you so much for trying it out! I'm so pleased! We have a separate [gem for logging](https://github.com/scoutapp/scout_apm_ruby_logging) that currently is focused on Rails. It seems like it should be possible to create an external logger that would transform the logs to our expected format and ship them to us via OTel SDK. I'll have to check it out later.

rsamoilov:
Hey @lancetarn,

Just a quick update - I added some load, tested a bit more with Active Record, and it looked great. There's no data if integrating Rage into a Rails app, but in this mode Rails instrumentation is being picked up, so maybe it's expected.

PR #614, opened 2026-03-26
Fix HTTP instrumentation for http gem v6.0+

### LINKED ISSUE #613: HTTP instrumentation incompatible with http gem v6.0
The HTTP client instrumentation in `ScoutApm::Instruments::HTTP` is incompatible with [http gem](https://github.com/httprb/http) v6.0+, which changed `HTTP::Client#request` from positional arguments to keyword arguments.

Error
```
ArgumentError: wrong number of arguments (given 3, expected 2)
```

Root Cause

Both the prepend and alias instrumentation paths in `lib/scout_apm/instruments/http.rb` pass opts as a third positional argument to `HTTP::Client#request`. In http v5, the method accepted `(verb, uri, opts = {})`. In v6, the signature changed to `(verb, uri, **opts)` - keyword arguments only - so the third positional argument raises an ArgumentError.

lancetarn:
## Summary

- Fixes #613 — `ArgumentError: wrong number of arguments (given 3, expected 2)` when using http gem v6.0+
- http gem v6.0 changed `HTTP::Client#request` from positional args `(verb, uri, opts = {})` to keyword args `(verb, uri, **opts)`
- Adds version detection via `Gem::Version` and separate instrumentation paths for v5 (positional) and v6+ (keyword) in both the prepend and alias code paths

## Test plan

- [x] Existing HTTP instrumentation test updated and passes with http gem v6.0.2
- [x] Full test suite (284 tests) passes with no regressions
- [x] Verify with http gem v5.x to confirm backwards compatibility

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,5 +1,7 @@
 # Pending
 
+- Fix compatibility with `http >= 6.0.0` (#613)
+
 # 6.1.1
 
 - Fix Redis 5 prepend interaction (#611)
--- a/lib/scout_apm/instruments/http.rb
+++ b/lib/scout_apm/instruments/http.rb
@@ -24,29 +24,62 @@ module ScoutApm
 
           if prepend
             ::HTTP::Client.send(:include, ScoutApm::Tracer)
-            ::HTTP::Client.send(:prepend, HTTPInstrumentationPrepend)
+            if http_version_at_least?("6.0.0")
+              ::HTTP::Client.send(:prepend, HTTPInstrumentationPrependV6)
+            else
+              ::HTTP::Client.send(:prepend, HTTPInstrumentationPrepend)
+            end
           else
-            ::HTTP::Client.class_eval do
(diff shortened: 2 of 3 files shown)


quinnmil (approved):

kjanoudi:
We hit the same incompatibility in production after upgrading to `http` 6.

Our environment:
- `scout_apm` 6.1.1
- `http` 6.0.2
- Ruby 3.4.7
- Rails 8.0.2.1

Observed error:
`ArgumentError: wrong number of arguments (given 3, expected 2)`

One failing path for us was a Sidekiq job calling `HTTP.get(..., params: ...)` through `IpInfo.fetch_ipstack`, but we also found other backend `HTTP` calls using `params:`, `json:`, and `form:` that would go through the same Scout wrapper.

We applied a local compatibility patch that forwards `**opts` in both the alias and prepend instrumentation paths, and that resolved the issue for us. So this PR matches what we needed downstream.

Thanks for putting this together.

PR #624, opened 2026-07-16
Use prepend with grape >= 3.3.0

jrothrock:
## Problem

Customer report: requests 500 with `#<SystemStackError: stack level too deep>` on scout_apm 6.2.0 + grape 3.3.1/3.3.2, with a stack alternating between:

```
grape (3.3.1) lib/grape/testing.rb:8:in 'Grape::Testing::RunBeforeEach#run'
scout_apm (6.2.0) lib/scout_apm/instruments/grape.rb:62:in run_with_scout_instruments
```

## Root cause

Grape 3.3.0 added `lib/grape/testing.rb` (eager-loaded via zeitwerk at `require 'grape'`), which does `Grape::Endpoint.prepend(Grape::Testing::RunBeforeEach)` — a `run` override that calls `super`.

Our instrument builds an alias-method chain on `Grape::Endpoint#run`. With that module prepended, `alias run_without_scout_instruments run` copies the **prepended module's** `run` rather than the endpoint's own method. The copy's `super` resolves to the current `Endpoint#run` — which is now our wrapper — so the two methods invoke each other until the stack overflows on the first request. Affects grape >= 3.3.0 with the default (alias) instrumentation method.

## Fix

- The Grape instrument now honors the `prepend:` install option (previously accepted but ignored), adding `GrapeEndpointInstrumentsPrepend` following the same pattern as the Net::HTTP/Redis5/Elasticsearch instruments. `use_prepend` / `prepend_instruments` config now works for Grape.
- Safety fallback: when `Grape::Endpoint.instance_method(:run).owner != Grape::Endpoint` (i.e. something is prepended over `#run`, as grape >= 3.3.0 does), force prepend mode even under default config, since the alias chain can only recurse in that case.
- Grape < 3.3.0 keeps the alias chain by default — no behavior change.

## Verification

Reproduced against a real grape 3.3.2 + puma app using this gem via `path:` — 500/SystemStackError before, 200s after, in all four combinations (grape 3.2.1/3.3.2 × alias/prepend), with identical captured layer names (`Grape/GET/HelloAPI/widgets/:id`) in both modes.

- New regression test `test/unit/instruments/grape_test.rb` actually runs an endpoint through Rack (fails with SystemStackError before this fix on grape >= 3.3.0), and asserts the chosen install method per grape version. Added `grape` to `gems/instruments.gemfile`.
- Full unit suite: 277 runs, 615 assertions, 0 failures, 0 errors on grape 3.3.2 and 3.2.1 (ruby 3.3.6).


jrothrock:
Pushed a fix for the Ruby 3.0 CI failures. Root cause: the matrix runs `gems/instruments.gemfile` on Ruby 3.0 with `SCOUT_USE_PREPEND=true`, where bundler resolves grape 3.0.1 and the instrument legitimately installs via prepend — but the test predicted alias vs. prepend from the grape version alone and refuted the prepend module in that job.

The test now accepts either install mechanism, and only *requires* prepend (with no alias chain present) when grape itself prepends over `Grape::Endpoint#run` (grape >= 3.3.0). It is also now gated on `SCOUT_TEST_FEATURES=instruments` like the other instrument tests, and grape was added to `gems/instruments-ruby4.gemfile` so the Ruby 4 instruments job covers it too.

Verified locally: Ruby 3.0 (grape 3.0.1), 3.3 (grape 3.0.1/3.2.1/3.3.2), and 4.0 (grape 3.3.2), each with and without `SCOUT_USE_PREPEND=true` — all green, plus full-suite runs on 3.0 and 3.3 in both modes.

mitchh456 (approved):

PR #635, opened 2026-09-22
Capture route names for Grape 4

### LINKED ISSUE #634: Support grape version 4 Endpoint#config
scout_apm builds `Grape/#{method}/#{api}/#{path}` from Grape 3 `options[:method|:path|:for]`.

Grape 4 moved those to `Endpoint#config`, so every request becomes `Grape/Unknown`.

Replicated in [v6.3.0](https://github.com/scoutapp/scout_apm_ruby/releases/tag/v6.3.0)

https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/instruments/grape.rb
https://github.com/ruby-grape/grape/blob/v4.0.1/UPGRADING.md

jrothrock:
## Summary

Grape 4 endpoints were all being reported as `Grape/Unknown` instead of their real route name (e.g. `Grape/GET/WidgetAPI/widgets/:id`), so every Grape transaction collapsed into one bucket in Scout.

Fixes #634.

### Root cause
The instrument built the transaction name from Grape 3 internals — `endpoint.options[:method]`, `options[:for]`, `options[:path]`. Grape 4 moved that route metadata off the public `options` Hash onto the (protected) `Endpoint#config` (`Grape::Endpoint::Options`), and renamed `options[:for]` → `endpoint.api`. Under Grape 4 those `options[...]` reads return `nil`, `.first` raises, and the rescue falls back to `Grape/Unknown`.

### Fix
Extract the name-building logic (previously duplicated across the alias-method and prepend wrappers) into a shared `GrapeEndpointNaming.name_for` helper that:
- reads `http_methods` / `api` / `path` from `config` on Grape 4, and
- falls back to the legacy `options[...]` reads on Grape 3,

preserving the existing name format and the rescue-to-`Grape/Unknown` fallback.

### Tests
Adds `test_endpoint_name_reflects_the_route`, which drives a request through an endpoint and asserts the captured `Controller` layer name is the real route name (and specifically **not** `Grape/Unknown`). It fails on the old code and passes with the fix.

### CI: json pin
Separately, json 3.0 (released 2026-09) removed the `quirks_mode` keyword that Grape 3's ActiveSupport JSON formatter still passes, which broke the Ruby 3.0 / `gems/instruments.gemfile` matrix cell on any JSON-rendering request (unrelated to this change, but it turns that cell red). Pinned `json < 3` in that gemfile to keep the Grape 3 cell green until it moves off Grape 3.

### Verification
Verified end-to-end against a real multi-endpoint Grape 4 app: booted it, hit `GET/POST/DELETE` routes (incl. path params and nested namespaces) over real HTTP, and confirmed the reported Scout checkin payload contained correct per-endpoint metrics **and** slow-transaction traces — each named `Grape/<METHOD>/<API>/<path>`, none `Grape/Unknown`.


quinnmil (approved):
🍇

